PR #1, opened 2019-12-04
Initialize

lkwdwrd:
This code is not 100% complete for what I would like, but the code itself is functional for the time being.

This is a combination of the client interface code with the environment specific packages included.

There are a couple breaking changes made here, this is a good opportunity since it is a new SDK package.

SearchV2 is included in this PR

Clients no longer have `read`, `write`, `update`, `delete` methods, instead they are renamed to `readRecord`, `writeRecord`, `updateRecord`, `deleteRecord`, etc. This makes naming in the SDK between records and notes look the same.

The crypto implementation abstraction and the way the final Tozny object are created are updated so that there is much less crypto code in the environment specific packages.

Babel is removed as a dependency for the main code base. Webpack is only used for the pre-compiled version. This will make working with this code so very much easier to work with. Better autocomplete support is included so that IDE and smart text editors are able to provide autocomplete hints, etc.

Some more updates that are minor. Testing is available for configured browers and node environment.

Documentation and internal comment updates are needed - this is just to get an initial commit into master.


README.md
```diff
@@ -1,2 +1,61 @@
...
+
+This repository contains interfaces and code providing the core of the Tozny platform JS SDKs. It defines types, API
+behaviors, and more to keep interactions consistent across the various code bases, yet allowing each SDK to define the cryptography layer and any context specific helpers for use in the target environment.
+
+## Use
+
+Import this package
+
+```sh
+npm install --save e3db-client-interface
```
galxy25: @lkwdwrd Is now not a good time to switch things over from the e3db to the Tozny namespace?

__tests__/environment/runner.js
```diff
@@ -0,0 +1,25 @@
...
+
+const TestRunner = require('jest-runner')
+
+const TestBrowser = process.env.TEST_BROWSER
+const TestEnvironment = process.env.TEST_ENVIRONMENT
+
+class BrowserRunner extends TestRunner {
+  constructor(...attr) {
+    super(...attr)
+    if (TestEnvironment !== 'remote' && TestBrowser === 'safari') {
```
galxy25: `&& (TestBrowser === 'safari' || (TestBrowser === 'edge' )`

browser/sodium/sodium.js
```diff
@@ -0,0 +1,158 @@
...
+    },
+    keyMaterial,
+    { name: 'HMAC', hash: { name: 'SHA-1' }, length: length * 8 },
+    true,
+    ['sign', 'verify']
+  )
+  return window.crypto.subtle.exportKey('raw', key)
+}
+
+function maybeToBytes(val) {
```
willmichael: ToBufferOrByteArray?

types/searchResult.js
```diff
@@ -0,0 +1,70 @@
...
+    if (!response.results) {
+      return []
+    }
+
+    /* eslint-disable */
+    let records = await Promise.all(
+      response.results.map(async result => {
+        const meta = await Meta.decode(result.meta)
+        const record = new Record(meta, result.record_data)
+        if (this.request.includeData) {
```
willmichael: needed?

types/searchResult.js
```diff
@@ -0,0 +1,70 @@
...
+        return record
+      })
+    )
+
+    this.request.nextToken = response.last_index
+    this.totalResults = response.total_results
+    if (response.search_id) {
+      this.searchId = response.search_id
+    }
+    if (records.length < this.request.count) {
```
willmichael: I *believe* because search service is eventually consistent this could lead to a situation where an un-share happens at search time where PDS will only return a subset of records returns from the ES search.  

The above last_index determination of done is more reliable, so I don't think this check is necessary
```
    if (response.last_index === 0) {
      this.done = true
    }
```

types/searchResult.js
```diff
@@ -0,0 +1,70 @@
...
+    this.done = false
+    this.totalResults = -1
+  }
+
+  /**
+   * Get the next page of results from the current query
+   *
+   * @returns {Promise<array>}
+   */
+  async next() {
```
willmichael: This is a really nice helper

README.md
lkwdwrd: yes, this whole readme has not been updated and this example code is incorrect. For the purposes of this PR, this is out of scope. I'm planning on updating this and much more during the upcoming hack days.

__tests__/environment/runner.js
lkwdwrd: Probably, but planning on making a better whitelist. This is a fine change for the time being.

browser/sodium/sodium.js
lkwdwrd: Only a buffer in node. Ensures a UInt8Array - this is an internal compatibility function.

types/searchResult.js
lkwdwrd: nope!

types/searchResult.js
lkwdwrd: I'll look at it, but I've run into bugs with poor server reporting of done -- this fixes it, but it is a bandaid. That was mostly with search v1, too.

lkwdwrd: OK -- here is my question for search V2 -- when you reach the last page of results, do you have to make a request that is out of range before you get back a last_index of 0?

which is correct?:

request next null:
  return 10 results and last index of 9
request next 9:
  return 4 results and last index of 13
request next 13:
  return 0 results and last index of 0

== OR ==

request next null:
  return 10 results and last index of 9
request next 9:
  return 4 results and last index of 0

PR #3, opened 2019-12-24
Update sdk for setting and parsing an identities first and last name

galxy25:


lib/identity/realm.js
```diff
@@ -108,10 +108,12 @@ class Realm extends CryptoConsumer {
    * @param {string} password The secret used to protect the users identity and encryption keys.
    * @param {string} token The registration token to create the storage client with.
    * @param {string} email The email address used for email brokered access to the identity.
+   * @param {string} firstName The first name to associate with the identity.
+   * @param {string} lastName The last name to associate with the identity.
    *
    * @return {Client} The identity Client for the user that was just registered with the realm.
    */
-  async register(username, password, token, email) {
+  async register(username, password, token, email, firstName, lastName) {
```
lkwdwrd: This method is getting a lot of parameters. I’ll approve for this PR, but we need to switch to an object param soon to allow for variable arguments.

PR #6, opened 2020-01-03
Don't send welcome emails with undefined for the identities name.

galxy25:


lib/identity/realm.js
```diff
@@ -212,26 +212,37 @@ class Realm extends CryptoConsumer {
...
+      eacp: {
+        email_eacp: {
+          email_address: email,
+          template: 'claim_account',
+          provider_link: this.brokerTargetUrl,
+          default_expiration_minutes: emailEACPExpiryMinutes,
+        },
+      },
+    }
+    if (firstName !== undefined || lastName !== undefined) {
```
lkwdwrd: This logic is a little messy. Rather than the nested if-statements, I'd rather detect undefined and switch to empty string instead. This could be done somewhat functionally like this:

```
const identityName = [firstName, lastName].filter(n => typeof n === 'string' && n !== '').join(' ');
```

_note this is not tested code and may not work perfectly!_

The logic is thus if it is unclear:

1. Turn first and last name into an iterable array
1. Filter out any non strings and empty values
1. join the filtered list with a space

If no names are sent, they are filtered out and an empty string is returned.

If only first name is sent, last name is filtered out and first name is returned on its own.

If only last name is sent, first name is filtered out and last name is returned on its own.

If both names are sent, they are joined by a space giving you `${first} ${last)`.

PR #7, opened 2020-02-10
Updates to support API style login

lkwdwrd:
Introduces a 'Partial Client' which represents a client that has their Tozny credentials but no Realm IDP App token, meaning they are only partially logged in. This also adjusts login to use API-style by default.


lib/identity/client.js
```diff
@@ -1,7 +1,6 @@
 const { validateResponseAsJSON, urlEncodeData } = require('../utils')
-const { deriveNoteCreds } = require('./shared')
 const fetch = require('isomorphic-fetch')
-const CryptoConsumer = require('../crypto/cryptoConsumer')
+const PartialClient = require('./paritalClient')
```
willmichael: paritalClient

lib/identity/realm.js
```diff
@@ -376,6 +462,8 @@ class Realm extends CryptoConsumer {
       this.apiUrl
     )
 
+    const configObject = storedCreds.data
+    configObject.agent = tokenInfo
     const user = this.fromObject(storedCreds.data)
```
willmichael: take config object

PR #13, opened 2020-03-24
Fix/email template

lkwdwrd:
Sends the email template when requesting an email reset for an identity.

Also
- Fixes a few bugs discovered along the way with various crypto methods
- Updates dependencies to alleviate some of the security problems found recently downstream


willmichael:
> easiest 1300 lines in a long time

agreed

PR #16, opened 2020-04-10
Feature/travis

lkwdwrd:
Modifications to automatically run tests using TavisCI when new commits are made. We'll have to see how this goes with our Sauce Labs account. It doesn't seem to take too much from our testing limit despite Travis taking a while to do the test run (~20 minutes for all environments). Each environment in Sauce Labs seems to take ~30s of test time.

Notice in the long list of commits here -- there is an indicator of when the tests pass or fail. You can see as I got to the working configuration :). When this is merged, a build-status badge is included in our Readme.


efabens:
Approved again on green

PR #18, opened 2020-06-18
Changes to support federated users, initially LDAP APIs

lkwdwrd:
These changes allow the login flow to support a federated flow. This includes:

- detecting a federated flow and auto-adjusting login behavior
- using the broker for federated logins, protected with a 'basic' TozID EACP
- automatically handling some login-actions so the consuming applications don't need to

The Identity Portal can consume a new version of the SDK, which initially is the only update it will need for federated identity support.


lib/identity/realm.js
```diff
@@ -250,6 +251,38 @@ class Realm extends CryptoConsumer {
     return idClient
   }
 
+  async registerFederatedUser(username, data, cryptoKeys, signingKeys) {
+    // eslint-disable-next-line
+    console.log(username, data) // flag if seen in CR
```
willmichael, line 256: unsure of what this comment means, or if this log should remain. suggest - more clarity in the comment

lib/identity/realm.js
```diff
@@ -250,6 +251,38 @@ class Realm extends CryptoConsumer {
...
+          public_signing_key: signingKeys.publicKey,
+          private_signing_key: signingKeys.privateKey,
+          api_url: this.apiUrl,
+        },
+      })
+      const brokerId = result.realm_broker_identity_tozny_id
+      await writeFederationNote(user, brokerId, true)
+    } catch (e) {
+      // eslint-disable-next-line
+      console.error(e) // flag if seen in CR
```
willmichael, line 282: same as above, we probably want to do more than console log if we're handling the error

lib/identity/realm.js
```diff
@@ -322,7 +358,35 @@ class Realm extends CryptoConsumer {
...
+          data = {
+            public_key: cryptoKeys.publicKey,
+            public_signing_key: signingKeys.publicKey,
+          }
+          break
+        case 'complete-broker-registration':
+          await this.registerFederatedUser(
+            username,
+            resp,
+            cryptoKeys,
```
willmichael, line 381: aren't cryptokeys and signing keys only valid if the above case is run?
```
        case 'register-brokered-user':
          cryptoKeys = await this.storage.generateKeypair()
          signingKeys = await this.storage.generateSigningKeypair()
```

lib/identity/realm.js
```diff
@@ -295,6 +328,8 @@ class Realm extends CryptoConsumer {
       }
     )
     let sessionStart = await credentialedDecodeResponse(request)
+    const { federated } = sessionStart
+    delete sessionStart.federated
```
willmichael, line 332: Why do we need to delete the federated flag?

lib/identity/realm.js
willmichael: Ah I think I'm understanding the federated login flow, for the future I think pulling it out into a separate method would help with clarity. (doesn't have to be done in this PR)

lib/identity/realm.js
```diff
@@ -354,18 +418,32 @@ class Realm extends CryptoConsumer {
...
-      this.storage.readNoteByName(
-        noteName,
-        cryptoKeyPair,
-        signingKeyPair,
-        {},
-        {
+    let storedCreds
+    if (federated) {
+      const keyNoteName = await this.crypto.hash(
+        `federated:${username}@realm:${this.realmName}`
```
willmichael, line 424: Ignore me if this is completely off, but would a user ever want to linked to multiple external idms? Do we need additional context in this naming scheme for later?

lib/identity/realm.js
efabens: curious about this as well

lib/identity/realm.js
```diff
@@ -573,30 +655,54 @@ class Realm extends CryptoConsumer {
...
+      parameters,
+      brokerType,
+      brokerUrl
+    )
+
+    const user = this.fromObject(storedCreds.data)
+
+    // Detect old format configuration. If found, migrate it.
+    // The config is a string here. Rather than parse it twice, check to see
+    // if the storage object contain a came case key string
```
efabens, line 679: camel case

lib/identity/realm.js
lkwdwrd: this is a quick insert I have in VS code, specifically used when something should _not_ go out into the world. Will remove these logs.

lib/identity/realm.js
lkwdwrd: The session data is pulling double duty now.. It's not my favorite way to do this, but was a way to get it done quickly. Later all of the items in `sessionStart` get turned into query parameters, 'federated' is not needed. This pulls it into a local var and then removes it from the object for query param usage.

This can and should be much better, but timelines...

lib/identity/realm.js
lkwdwrd: yeah, it's a bit odd -- the way the required action is written these will always be in series. This could be an intermediate object, or something, but we're trying to get this out the door, so I just went with higher scoped vars to pass the data from one action to the next. I'll add some code comments to that effect.

lib/identity/realm.js
lkwdwrd: so... somewhat no, that's not how Keycloak functions.. A user has a single federation link. _However_ this one note will work for any provider where Keycloak becomes the primary authentication factor. So if you were connected to two different LDAPs, or something, authenticating with either one would be enough to allow Keycloak to issue your session token, which could be used to request this broker key note from the hosted broker. Getting a valid session token proves you've authenticated with Keycloak, which in a federated world, is enough for the broker to release the credentials. It is important that in non-federated accounts this doesn't happen.

lib/identity/realm.js
lkwdwrd: I've added some comments to hopefully clarify what is happening a bit. This is definitely a sprawling method that could use some sub-functions to clean up the logic.

lib/identity/realm.js
lkwdwrd: comment to explain what is happening and suggest better structure down the road.

PR #21, opened 2020-09-01
Update to support Realm Domain and Realm Name

efabens:
- Use realm domain as key for identity paths
- User realm name for generating identity notes


lib/identity/realm.js
```diff
@@ -433,7 +441,7 @@ class Realm extends CryptoConsumer {
     let storedCreds
     if (federated) {
       const keyNoteName = await this.crypto.hash(
-        `federated:${username}@realm:${this.realmName}`
```
willmichael, line 436: So we derive this note with realm domain but other notes with realm name?

PR #26, opened 2020-12-18
create group support, plus test

carrala:
- create group support with test that passes
- read, list, and delete support and testing are coming


lib/storage/client.js
```diff
@@ -1044,6 +1044,32 @@ class Client extends CryptoConsumer {
     const params = { id_string: noteName }
     return shared.issueNoteChallenge(this, params, meta)
   }
+
+  /**
+   * createGroup is a static method that creates a groups and sends it to TozStore,
```
efabens, line 1049: I don't believe this is a static method given it's reference to `this`

lib/storage/shared.js
```diff
@@ -487,6 +490,54 @@ async function issueNoteChallenge(client, requestParams, body = {}) {
   return request.json()
 }
 
+async function createGroupRequest(
+  crypto,
+  data,
+  encryptionKeyPair,
+  recipientEncryptionKey
```
efabens, line 497: I don't believe this field is needed. The function should generate the new group keypair

lib/storage/shared.js
```diff
@@ -487,6 +490,54 @@ async function issueNoteChallenge(client, requestParams, body = {}) {
...
+) {
+  const accessKey = await crypto.randomKey()
```
efabens, line 499: This should be a keypair and not just a single key

efabens: https://github.com/tozny/e3db-clients-go/blob/e0e19af0ad0c22931e2aca4b3ea6ea7fb09736d6/storageClient%2FstorageClient_test.go#L280 is a good reference

lib/storage/client.js
```diff
@@ -1044,6 +1044,32 @@ class Client extends CryptoConsumer {
...
+   * allowing you to supply your own encryptionKey pair.
+   *
+   *
+   * @param {object} data  A hashmap of the data in the group - for now this just includes groupName
+   * @param {string} recipientEncryptionKey The public encryption key of the reader of this note
```
efabens, line 1054: This is not the correct doc

lib/storage/client.js
```diff
@@ -1044,6 +1044,32 @@ class Client extends CryptoConsumer {
     const params = { id_string: noteName }
     return shared.issueNoteChallenge(this, params, meta)
   }
+
+  /**
+   * createGroup is a static method that creates a groups and sends it to TozStore,
+   * allowing you to supply your own encryptionKey pair.
+   *
+   *
+   * @param {object} data  A hashmap of the data in the group - for now this just includes groupName
```
lkwdwrd, line 1053: these data objects should get created as types if they are structured

lib/storage/client.js
```diff
@@ -1044,6 +1044,32 @@ class Client extends CryptoConsumer {
...
+   * @param {string} recipientEncryptionKey The public encryption key of the reader of this note
+   *
+   * @returns {Group} A response from TozStore; the group that has been written.
+   */
+  async createGroup(data, recipientEncryptionKey) {
```
lkwdwrd, line 1058: what is the `recipientEncryptionKey` here? Ganearlly in SDK method's we'll refer to clients by client ID, but in createGroup I would think the client would add themselves as the 'recipient' in which case we could omit this parameter, right?

lib/storage/client.js
lkwdwrd: ah -- looks like it is based on the shared method below 👍 . Let's call this something more like `groupData` and indicate the type here instead of 'object' in the doc-block.

carrala:
I think these changes make this more like what it is supposed to be.

I also noticed that the CI check failed after the first commit, so if anyone has any tips on what I need to do for that it would be much appreciated.

lib/storage/client.js
```diff
@@ -1044,6 +1044,30 @@ class Client extends CryptoConsumer {
     const params = { id_string: noteName }
     return shared.issueNoteChallenge(this, params, meta)
   }
+
+  /**
+   * createGroup creates a group and sends it to TozStore.
+   *
+   * @param {string} name  the name of the group
+   * @param {array} capability an array of strings of the capabilities the group will have
```
efabens, line 1052: Perhaps `capabilities` given it is an array

lib/storage/shared.js
```diff
@@ -487,6 +490,53 @@ async function issueNoteChallenge(client, requestParams, body = {}) {
...
+    encryptionKeyPair
+  )
+  let groupData = new GroupData(name, capabilities)
+  let groupKeys = new GroupKeys(encryptionKeyPair.publicKey, eak)
+  let group = new Group(groupData, groupKeys)
+  return group
+}
+
+/*
+ * InternalCreateGroup is an internal method used by createGroup
```
efabens, line 511: should this match the function name?

efabens (approved):

lib/crypto/crypto.js
```diff
@@ -257,6 +257,32 @@ class Crypto {
...
+   * Encrypt a private key using a keypair.
+   *
+   * @param {string} privateKey
+   * @param {KeyPair} encryptionKeyPair
+   *
+   * @return {Promise<string>}
+   */
+  async encryptPrivateKey(privateKey, encryptionKeyPair) {
+    let rawPrivateKey = this.platform.b64URLDecode(privateKey)
+    let publicEncKey = this.platform.b64URLDecode(encryptionKeyPair.publicKey)
```
lkwdwrd, line 270: Do we do this in the encrypt eak? A lot of times you are going to have a private key and public key from different key pairs, so its probably a good idea to split them into to separate parameters. (example: If I write a membership key for your client, I'll use my private key, and your public key).

lib/storage/client.js
```diff
@@ -1044,6 +1044,30 @@ class Client extends CryptoConsumer {
...
+   *
+   * @returns {Group} A response from TozStore; the group that has been written.
+   */
+  async createGroup(name, capabilities = []) {
+    /* eslint-enable */
+    const encryptionKeys = new KeyPair(
+      this.config.publicKey,
+      this.config.privateKey
+    )
+    return shared.createGroup(
```
lkwdwrd, line 1062: I didn't think of this before, but I am a bit curious why this ended up in the shared functions? You may have been following patters, but in general shared is for methods that are re-used accross objects or different contexts. So for instance with notes, there is 'anonymous' notes which do not require a client, and 'regular' notes which are written by a client. The code goes in shared so the code to write a note is the same no matter if it is written from the anonymous or regular contexts. I don't think we'll ever be creating a group anonymously, or allow this to run in an object other than a client. In that case we can skip shared and just define the behavior here in the client.

types/group.js
```diff
@@ -0,0 +1,62 @@
+const GroupKeys = require('./groupKeys')
+const Serializable = require('./serializable')
+
+class Group extends Serializable {
+  constructor(data, groupKeys) {
```
lkwdwrd, line 5: I'm not quite in line with the types we have defined here. I think we're going to run into some problems going forward and will be forced to refactor later.

Let's try to split this up a bit and think about what concept each type maps to (not saying this is definitely the way to go, just what is making sense in my mind as I think this through).

- **Group**: When I think of a `Group` object, I think of the data returned from a `GroupInfo()` request. Name, public key, group ID, possibly some other metadata. This will be generic no matter which group member is fetching it.
- **Membership(Meta?)**: This could be a user's connection to a group -- containing their group capabilities, their membership key definition, other group-client metadata.
- **MembershipKey**: (Basically what you have for GroupKeys?). This could be a handy object, though I suggest we call it `MembershipKey since that we what we've called the encrypted group private key?

This is not much different the the types you have, but conceptually I'm not sure a Group should take a Membership as part of it's constructor -- Perhaps it is the Membership type which contains a Group instead, to make easy access to Group info from a client's Membership to that group?

Plenty open to other thought's opinions here.

lkwdwrd:
https://travis-ci.org/github/tozny/js-sdk/builds/750463854 -- Node tests are passing, browser tests are failing. It is saying client.createGroup is not a function. These use a built-version of the SDK to run. Try running `npm run build` and check https://github.com/tozny/js-sdk/blob/master/DEVELOPMENT.md#testing-with-travisci if that still doesn't work. See https://github.com/tozny/js-sdk/blob/master/DEVELOPMENT.md#building for build instructions)

types/group.js
```diff
@@ -0,0 +1,58 @@
+const GroupMembershipKeys = require('./groupMembershipKeys')
+const Serializable = require('./serializable')
+
+class Group extends Serializable {
+  constructor(data, membership, membershipKeys) {
+    super()
+    this.groupName = data.groupName
+    this.publicKey = membershipKeys.publicKey
+    this.encryptedGroupKey = membershipKeys.encryptedGroupKey
+    if (membership !== null) {
```
galxy25, line 10: Whats the thinking behind this if?

types/membership.js
```diff
@@ -0,0 +1,35 @@
+const { GroupData } = require('./groupData')
+const Serializable = require('./serializable')
+
+class Membership extends Serializable {
```
galxy25, line 4: Prefer GroupMembership for this class so its clear its a type representing a member of a group.

galxy25 (approved):

lib/crypto/crypto.js
```diff
@@ -257,6 +257,33 @@ class Crypto {
...
+   *
+   * @return {Promise<string>}
+   */
+  async encryptPrivateKey(privateKey, writerPrivateKey, readerPublicKey) {
+    let rawPrivateKey = this.platform.b64URLDecode(privateKey)
+    let publicEncKey = this.platform.b64URLDecode(readerPublicKey)
+    let privateEncKey = this.platform.b64URLDecode(writerPrivateKey)
+
+    let nonce = await this.randomNonce()
+    let eak = await this.provider.encryptAsymmetric(
```
lkwdwrd, line 275: nit: this is not an eak (encrypted access key), it's a private key, so let's update the var name to something like 'pk'.

lib/storage/client.js
```diff
@@ -1044,6 +1047,65 @@ class Client extends CryptoConsumer {
...
+   * Make a request object for the create group request
+   * @param {string} name
+   * @param {Array} capabilities
+   * @param {KeyPair} encryptionKeyPair
+   *
+   * @return {Group}
+   */
+  async createGroupRequest(name, capabilities = [], encryptionKeyPair) {
+    const groupKeyPair = await this.crypto.generateKeypair()
+    const eak = await this.crypto.encryptPrivateKey(
```
lkwdwrd, line 1061: nit: same as above, let's call this pk or or similar to indicate it is a private key, not an eak

lib/storage/client.js
```diff
@@ -1044,6 +1047,65 @@ class Client extends CryptoConsumer {
...
+      groupKeyPair.privateKey,
+      encryptionKeyPair.privateKey,
+      encryptionKeyPair.publicKey
+    )
+    let groupData = new GroupData(name, capabilities)
+    let membershipKeys = new GroupMembershipKeys(
+      encryptionKeyPair.publicKey,
+      eak
+    )
+    let group = new Group(groupData, null, membershipKeys)
```
lkwdwrd, line 1071: will call this out below in the type definitions, but it's a small code smell to see null passed to a constructor with any regularity. It generally means there is an improvement that can be made in the object composition.

lib/storage/client.js
```diff
@@ -1044,6 +1047,65 @@ class Client extends CryptoConsumer {
...
+
+  /**
+   * Make a request object for the create group request
+   * @param {string} name
+   * @param {Array} capabilities
+   * @param {KeyPair} encryptionKeyPair
+   *
+   * @return {Group}
+   */
+  async createGroupRequest(name, capabilities = [], encryptionKeyPair) {
```
lkwdwrd, line 1059: nit: two things to consider:
- does this method need to be public?
- does this method need to be async?

one additional question:
- does this method belong here?

This last one probably warrants a real-time discussion real quick :)

lib/storage/client.js
```diff
@@ -1044,6 +1047,65 @@ class Client extends CryptoConsumer {
...
+  /**
+   * createGroup creates a group and sends it to TozStore.
+   *
+   * @param {string} name  the name of the group
+   * @param {array} capabilities an array of strings of the capabilities the group will have
+   *
+   * @returns {Group} A response from TozStore; the group that has been written.
+   */
+  async createGroup(name, capabilities = []) {
+    // /* eslint-enable */
```
lkwdwrd, line 1084: can this get removed? Why is there a need to enable this here?

lib/storage/client.js
```diff
@@ -1044,6 +1047,65 @@ class Client extends CryptoConsumer {
...
+    const encryptionKeyPair = new KeyPair(
```
lkwdwrd, line 1085: could very well be no, but does this generation step belong in the `createGroupRequest()` method?

lib/storage/client.js
```diff
@@ -1044,6 +1047,65 @@ class Client extends CryptoConsumer {
...
+      `${this.config.apiUrl}/v2/storage/groups`,
+      {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+        body: createGroupReq.stringify(),
+      }
+    )
+    let storedGroupResp = await checkStatus(response)
```
lkwdwrd, line 1104: there is an additional helper which checks status and decodes to JSON. Not all methods use it because it is a somewhat newer addition to utilities, but it will clean up some intermediate vars here.

lib/storage/client.js
```diff
@@ -1044,6 +1047,65 @@ class Client extends CryptoConsumer {
...
+   *
+   * @returns {Group} A response from TozStore; the group that has been written.
+   */
+  async createGroup(name, capabilities = []) {
+    // /* eslint-enable */
+    const encryptionKeyPair = new KeyPair(
+      this.config.publicKey,
+      this.config.privateKey
+    )
+    let createGroupReq = await this.createGroupRequest(
```
lkwdwrd, line 1089: super nit: While much code in here uses let, I prefer const unless the value needs to change. This is purely an opinion. I'll take a larger look at this file to see if it is fully consistent with let or mixed.

types/group.js
```diff
@@ -0,0 +1,58 @@
+const GroupMembershipKeys = require('./groupMembershipKeys')
+const Serializable = require('./serializable')
+
+class Group extends Serializable {
```
lkwdwrd, line 4: We should adjust object composition here. `Membership` should hold a `Group` instead of the other way around. My thinking here is you do not have a membership without a group, but you can have a group definition without an active membership.

This said, your head is in this much more than mine is, I may be overlooking a reason why this doesn't work.

lkwdwrd (changes requested):
While I've added a number of comments, they really are very small things. This is looking great from a macro level. Nice work!

Requesting changes just because we should have a short real-time discussion about a couple of things to make sure I'm not overlooking something in my suggestions before asking for them.

lkwdwrd:
Excellent real-time discussion with @carrala! We talked more through types and have decided on some re-organization and composition updates. Final types look something like:

* **Group**
  * Name (string)
  * ID (UUID, string)
  * Account ID (UUID, string)
  * Public Key (B64URL, string)
  * Created At (Date)
  * Modified At (Date)

* **Group Membership**
  * Client ID (UUID, string)
  * Group (Group)
  * Capabilities (Group Capabilities)

* **Membership Key**
  * Key Material (B64URL, string)
  * Authorizer Public Key (B64URL, string)
  * Authorizer ID (UUID, string)

* **Group Capabilities**
  * Manage Membership (boolean)
  * Read Content (boolean)
  * Share Content (boolean)

----

These types strike a nice balance of utility and user friendliness. Specifically the capabilities type helps ensure the user does not mis-type a string to define group capabilities. Excited to see a solid foundation to build from really shaping up.

carrala:
- create group returns groupMembership instead of just group
- new capabilities type lets users specify capabilities with booleans

galxy25 (approved):

efabens (approved):

PR #27, opened 2020-12-22
Get group by group_id & list groups

carrala:
Tests pass for both these parts.


__tests__/storage.test.js
```diff
@@ -213,13 +213,46 @@ describe('Tozny', () => {
     const deleted = await ops.deleteAnonymousNote(written.noteId, signingPair)
     expect(deleted).toBe(true)
   })
-  it('can create groups', async () => {
+  it('can create, read and list groups', async () => {
```
galxy25, line 216: Make each of these a separate test, generally we want each test to be scoped to testing a single higher level concept, that way if the test fails its clear exactly what functionality of the api is broken / has regressed.

galxy25 (approved):
Approved pending splitting out of mega test into individual test cases

lib/storage/client.js
```diff
@@ -1106,6 +1107,69 @@ class Client extends CryptoConsumer {
...
+    let request = await this.authenticator.tsv1Fetch(
+      `${this.config.apiUrl}/v2/storage/groups/${groupID}`,
+      {
+        method: 'GET',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      }
+    )
+    let storedGroup = await checkStatus(request)
```
lkwdwrd, line 1128: nit: we can use the helper which combines the status check/json decode here

lib/storage/client.js
```diff
@@ -1106,6 +1107,69 @@ class Client extends CryptoConsumer {
...
+    nextToken = null,
+    max = null
+  ) {
+    const urlData = {
+      client_id: clientID,
+      group_names: groupNames,
+      nextToken: nextToken,
+      max: max,
+    }
+    let encodedUrl = urlEncodeData(urlData)
```
lkwdwrd, line 1155: huh... I think this helper was actually written to url-encode POST data in request bodies, but this is a great use for it. Nice!

types/group.js
```diff
@@ -6,7 +6,9 @@ class Group extends Serializable {
     super()
     this.groupName = data.groupName
     this.publicKey = membershipKeys.publicKey
-    this.encryptedGroupKey = membershipKeys.encryptedGroupKey
+    if (membershipKeys.encryptedGroupKey !== null) {
```
lkwdwrd, line 9: is there a reason to not allow this as a null value?

lib/storage/client.js
```diff
@@ -1106,6 +1107,69 @@ class Client extends CryptoConsumer {
...
+    let request = await this.authenticator.tsv1Fetch(
+      `${this.config.apiUrl}/v2/storage/groups?${encodedUrl}`,
+      {
+        method: 'GET',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      }
+    )
+    let storedGroups = await checkStatus(request)
```
lkwdwrd, line 1165: combined helper again :)

lkwdwrd (approved):
Some very minor nits to consider and a question. This looks good!

PR #31, opened 2020-12-31
fix for list groups

carrala:
* handles the case where no groups are returned, which was previously causing an error.


lib/storage/client.js
```diff
@@ -1176,6 +1176,9 @@ class Client extends CryptoConsumer {
       }
     )
     let groupsJson = await validateResponseAsJSON(response)
+    if (groupsJson.groups === null) {
```
galxy25, line 1179: @carrala @rosasck this is something that should be fixed on the server side as well, unless there is a non 200 level error the api should always return a valid list groups response, even if its empty like below
```json
{
groups: []
}
```
}

galxy25 (approved):

efabens (approved):

lkwdwrd (approved):

carrala:
includes another fix for group names being added to the query string

galxy25 (approved):

efabens (approved):

PR #32, opened 2021-01-07
Add group members

rosasck:
no test yet


lib/storage/client.js
```diff
@@ -1182,6 +1183,64 @@ class Client extends CryptoConsumer {
...
+   * @param {string} groupId  UUID assigned by TozStore, used to identify a group.
+   * @param {GroupMember[]} groupMembers wraps together the ClientID, Capability and Membership Key for each client being added to the group
+   *
+   *
+   * @returns {GroupMember[]} A response from TozStore; the group members who have been added.
+   *
+   */
+  async addGroupMembers(groupId, groupMembers) {
+    if (!(groupMembers instanceof GroupMember)) {
+      throw new Error('Need a ')
```
carrala: I think you forgot to finish out this sentence.

__tests__/utils/operations.js
```diff
@@ -545,4 +545,17 @@ module.exports = {
     )
     return groupsJson
   },
+  async addGroupMembers(config, groupMembers) {
+    const result = await runInEnvironment(
+      function(configJSON, groupMembersJSON) {
+        var config = Tozny.storage.Config.fromObject(configJSON)
+        var client = new Tozny.storage.Client(config)
+        groupMembers = new Tozny.types.GroupMember(groupMembersJSON)
```
carrala, line 553: is groupMember able to parse an array of groupMembers correctly? my understanding was that it only sets up one groupMember at a time; however, i haven't run your code yet, so it's very possible that this does actually work.

rosasck: Currently attempting to test that, i think you are right tho!

carrala: I had a similar issue with listGroups. I think you can do something similar to what you're doing to decode the members in client.js (line 1229).

lib/storage/client.js
```diff
@@ -1182,6 +1183,64 @@ class Client extends CryptoConsumer {
...
+   *
+   * @param {string} groupId  UUID assigned by TozStore, used to identify a group.
+   * @param {GroupMember[]} groupMembers wraps together the ClientID, Capability and Membership Key for each client being added to the group
+   *
+   *
+   * @returns {GroupMember[]} A response from TozStore; the group members who have been added.
+   *
+   */
+  async addGroupMembers(groupId, groupMembers) {
+    if (!(groupMembers instanceof GroupMember)) {
```
galxy25, line 1198: I applaud the effort, but its javascript so better to embrace its untyped nature and not sporadically type check things (e.g. we're not type checking the groupId but we are for group members). Also I think this will fail / doesn't lined up with the documentation as the docs and variable name indicate it can be more than one group member, in which case instanceof == Array and that would cause this method to fail.

package.json
```diff
@@ -55,6 +55,7 @@
     "lint-staged": "^9.5.0",
     "prettier": "^1.19.1",
     "selenium-webdriver": "^4.0.0-alpha.7",
+    "typescript": "^4.1.3",
```
galxy25, line 58: Is this being used?

lib/storage/client.js
```diff
@@ -1182,6 +1183,64 @@ class Client extends CryptoConsumer {
...
+        body: JSON.stringify(addGroupMembersRequest),
+      }
+    )
+    // Await the Response from the API with the Added Group member Response
+    let groupMemberJson = await validateResponseAsJSON(response)
+    let listGroupMembers = []
+    groupMemberJson.forEach(groupMember => {
+      listGroupMembers.push(GroupMember.decode(groupMember))
+    })
+    switch (response.status) {
```
galxy25, line 1232: Take a look at what `validateResponseAsJSON` does, I think that because it catches any non 2xx response and throws an exception that you can and should remove the switch statement as you're only going to make it past line 1227 if the response is a 2xx level. (Which obviously I'm opposed to because it means that we've given up control to a function we call unless we want to wrap and catch the exception ourself, but as with everything, tradeoffs and time make pragmatists of us all, so it is what it is and we can work around it)

galxy25 (approved):
approved pending passing test

package.json
rosasck: ran npm install and that changed

rosasck:
Reworked to create membership key

rosasck:
right now its functional for a non bulk endpoint! Tests pass. Will work on making it a bulk endpoint next

lib/storage/client.js
```diff
@@ -1182,6 +1182,85 @@ class Client extends CryptoConsumer {
...
+        method: 'GET',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      }
+    )
+    let groupMembershipJson = await validateResponseAsJSON(membershipResponse)
+    // Get All members Public Keys to create their membership keys
+    let publicKeysResponse = await this.authenticator.tokenFetch(
+      `${this.config.apiUrl}/v1/client/${groupMembers.clientID}/public`,
```
willmichael, line 1210: This method will likely fail when done in batch -- The batch call is a POST endpoint living at /v1/client/public and it accepts:
```
type batchPublicRequest struct {
	ClientIDs []uuid.UUID `json:"client_ids"`
}
```

types/groupMember.js
```diff
@@ -0,0 +1,38 @@
+const Capabilities = require('./capabilities')
+const Serializable = require('./serializable')
+
+class GroupMember extends Serializable {
+  constructor(clientID, capabilities) {
+    super()
+    this.clientID = clientID
+    this.membershipKey = null
```
willmichael, line 8: Any reason to not init the membership key here?

lib/storage/client.js
```diff
@@ -1182,6 +1182,85 @@ class Client extends CryptoConsumer {
...
+        method: 'GET',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      }
+    )
+    let publicKeysJSON = await validateResponseAsJSON(publicKeysResponse)
+    // GET THE PUBLIC KEY FOR THE AUTHORIZER
+    let publicKeysAuthResp = await this.authenticator.tokenFetch(
+      `${this.config.apiUrl}/v1/client/${groupMembershipJson.authorizer_id}/public`,
```
willmichael, line 1221: Who is the authorizer in this scenario and why do we need their information here, isn't the caller of this method the one adding new group members and thus the authorizer for these clients?

willmichael (commented):
Added some comments to aid in adding the bulk endpoints, its important to note that at the end of the day the endpoint really just accepts an encrypted payload and it cannot do validation of the input, so additional verification is needed perhaps with other SDKS to verify the addition of the group member is validated -- We shouldn't merge this into master as is

rosasck:
Remove group members bulk endpoint is complete!, working on making add group member bulk now

lib/storage/client.js
rosasck: we need to know who wrote the membership key for the person adding these people. Since we need their public key in order to unwrap the membership key

lib/storage/client.js
rosasck: i couldnt get the array to work so i opted to do this to atleast get some functionality working!, but fixing that now

rosasck:
Add group member bulk endpoint complete  + tests passing

rosasck:
list group member and tests

rosasck:
list shared records with group is complete + test, i did have to do a small work around until i get the PR in storage service merged on lines 
1387-1389 in client.js 
this will be removed once that gets merged

rosasck:
JS SDK Completed Support:
- Add Group Member
- Remove Group Member
- List Group members
- List all Records shared with group

galxy25:
> JS SDK Completed Support:
> 
> * Add Group Member
> * Remove Group Member
> * List Group members
> * List all Records shared with group
> 
> Todo:
> 
> * Share Record with Group

@rosasck Excellent work, and reminder to merge early and often, this may be a good point to get this current pr merged and have a smaller one for faster review for the share record with group task

rosasck:
JS SDK Completed Support:

Add Group Member
Remove Group Member
List Group members
List all Records shared with group
Share Record with Group

__tests__/storage.test.js
```diff
@@ -291,4 +292,125 @@ describe('Tozny', () => {
...
+    let shareWithGroup = await ops.shareRecordWithGroup(
+      readerClient,
+      created.group.groupID,
+      type
+    )
+    let recordExpected = {
+      record_type: type,
+    }
+    expect(shareWithGroup).toMatchObject(recordExpected)
+  })
```
galxy25, line 415: this is some elegant test code. If you don't have to debug it, javascript can be quite pleasant to read at times 🤔

lib/storage/client.js
```diff
@@ -1182,6 +1182,251 @@ class Client extends CryptoConsumer {
...
+    // Create the new membership key
+    // for loop
+    for (var i = 0; i < groupMembers.length; i++) {
+      let newMembershipKey = await this.crypto.createGroupMembershipKey(
+        this.config.privateKey,
+        groupMembershipJson.encrypted_membership_key,
+        publicKeysJSON.clients[groupMembers[i].clientID].public_key[
+          'curve25519'
+        ],
+        publicKeysAuthJSON.public_key['curve25519']
```
galxy25, line 1249: These are magic and crucial values that should be extracted out to named constant, i.e. https://github.com/tozny/e3db-clients-go/blob/master/crypto.go#L22

galxy25 (approved):
👏🏽

rosasck:
Travis is failing since my storage service is only deployed to dev!

lib/crypto/crypto.js
```diff
@@ -717,6 +736,36 @@ class Crypto {
     transformer.transform(source)
     return destination.getReader()
   }
+  /**
+   * createGroupMembershipKey takes the encrypted membership key of the current group member
+   * and decrypts it, in order to encrypt it for the new group member.
+   *
+   * @param {stirng} groupMemberClientID
```
willmichael, line 743: string

lib/storage/client.js
```diff
@@ -1182,6 +1183,267 @@ class Client extends CryptoConsumer {
...
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify(groupMembersPublicKeysRequest),
+      }
+    )
+    let publicKeysJSON = await validateResponseAsJSON(publicKeysResponse)
+    // Create the new membership key
+    // for loop
```
willmichael, line 1242: do we need this comment?

lib/storage/client.js
```diff
@@ -1182,6 +1183,267 @@ class Client extends CryptoConsumer {
...
+      }
+    )
+    // Check the response code
+    switch (response.status) {
+      case 204:
+        return true
+      case 403:
+        throw new Error('Unauthorized')
+      default:
+        throw new Error('Error while deleting record data.')
```
willmichael, line 1327: Useful to add the status code to of response to provide some context of the failure, also we're "removing group members" here

lib/storage/client.js
```diff
@@ -1182,6 +1183,267 @@ class Client extends CryptoConsumer {
...
+      {
+        method: 'GET',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      }
+    )
+    let groupMembersJson = await validateResponseAsJSON(response)
+    let listGroupMembers = []
+    groupMembersJson.forEach(member => {
```
willmichael, line 1350: Is this necessary, aren't you just iterating over a list and pushing contents onto another list?

lib/storage/client.js
```diff
@@ -1182,6 +1183,267 @@ class Client extends CryptoConsumer {
...
+   * @param {string} groupId   UUID assigned by TozStore, used to identify a group.
+   * @param {string} recordType Record type to be shared with the group.
+   *
+   * @return {Promise<Record>}
+   */
+  async shareRecordWithGroup(groupId, recordType) {
+    // Get the Groups Public key
+    let groupInfo = this.readGroup(groupId)
+    // Get the encrypted Access Key
+    let encryptedAccessKey = await this.getReaderKey(
```
willmichael, line 1423: Will this method also create a reader key if it doesn't exist?

lib/storage/client.js
```diff
@@ -1182,6 +1183,267 @@ class Client extends CryptoConsumer {
...
+      `${this.config.apiUrl}/v2/storage/groups/${groupId}/share`,
+      {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify(shareRecordRequest),
+      }
+    )
+    let recordsJson = await validateResponseAsJSON(response)
```
willmichael, line 1445: recordsJson doesn't reflect what's being returned here, also the assignment is unecessary

types/groupMember.js
willmichael: Sorry I forget, did we have a discussion outside of why this shouldn't be part of the init?

__tests__/storage.test.js
```diff
@@ -291,4 +292,125 @@ describe('Tozny', () => {
...
+    const created = await ops.createGroup(writerClient, groupName)
+    const groupMember = new GroupMember(readerClient.clientId, { read: true })
+    let groupMembersToAdd = []
+    groupMembersToAdd.push(groupMember)
+    await ops.addGroupMembers(
+      writerClient,
+      created.group.groupID,
+      groupMembersToAdd
+    )
+    await ops.listGroupMembers(writerClient, created.group.groupID)
```
willmichael, line 377: The results of this call seem to be unused?

__tests__/storage.test.js
```diff
@@ -291,4 +292,125 @@ describe('Tozny', () => {
...
+    let sharedWithGroup = await ops.listRecordsSharedWithGroup(
+      readerClient,
+      created.group.groupID,
+      [],
+      0,
+      10
+    )
+    let sharedWithGroupExpected = []
+    expect(sharedWithGroup).toMatchObject(sharedWithGroupExpected)
```
willmichael, line 386: How are we identifying that records are listed successfully if we haven't written any, unless I'm missing them in this test?

__tests__/storage.test.js
```diff
@@ -291,4 +292,125 @@ describe('Tozny', () => {
...
+    await ops.writeRecord(readerClient, type, data, meta)
+    let shareWithGroup = await ops.shareRecordWithGroup(
+      readerClient,
+      created.group.groupID,
+      type
+    )
+    let recordExpected = {
+      record_type: type,
+    }
+    expect(shareWithGroup).toMatchObject(recordExpected)
```
willmichael, line 414: The share operation was successful, but how do we know that records are actually shared without listing records?

willmichael (commented):
Added some comment on comments, and questions about the tests

lib/storage/client.js
```diff
@@ -1182,6 +1183,267 @@ class Client extends CryptoConsumer {
...
+      let recordData = await new RecordData(recordsJson.results[i].record_data)
+      let record = await new Record(
+        meta,
+        recordData,
+        recordsJson.results[i].rec_sig
+      )
+      let decryptedRecord = shared.decryptRecord(this, record)
+      listSharedWithGroups.push(decryptedRecord)
+    }
+    return listSharedWithGroups
```
willmichael, line 1408: This methods needs to return the next token, so the caller of this method can paginate through his results

lib/storage/client.js
rosasck: nope

PR #38, opened 2021-01-25
Adding a group description

rosasck:
Cant be pushed until storage service is deployed for this to work, will fail tests until then


lib/storage/client.js
```diff
@@ -1055,10 +1055,11 @@ class Client extends CryptoConsumer {
    *
    * @param {string} name  the name of the group
    * @param {array} capabilities an array of strings of the capabilities the group will have
+   * @param {string} description  details about the group
    *
    * @returns {Group} A response from TozStore; the group that has been written.
    */
-  async createGroup(name, capabilities = []) {
+  async createGroup(name, capabilities = [], description) {
```
efabens, line 1062: We might want description to have a sane default. either an empty string or "A TozStore group"

rosasck:
added null as a place holder for description and changed json tag from group_description to description

lkwdwrd (approved):
Code looks good, but I see for some reason tests are failing when run in the Windows environment. When groups is fully ready and merged down, before we cut the new SDK release lets make sure the tests are passing first.

willmichael (approved):
Change looks simple enough, but do we need to test that group descriptions are returned when we access groups? (ignore me if i missed that in the tests)

PR #40, opened 2021-02-04
Create secrets with identity client

carrala:


lib/identity/client.js
```diff
@@ -67,6 +68,60 @@ class Client extends PartialClient {
...
+      }
+      var member = new GroupMember(this.storage.config.clientId, capabilities)
+      groupMembers.push(member)
+      // add read and share capabilities for the client
+      await this.storage.addGroupMembers(group.groupID, groupMembers)
+    } else {
+      group = groups.groups[0]
+    }
+    // create the record
+    const recordType = `tozny.secret.ae41d9cc-cd5c-4a90-8efe-0046784e195c.${secret.secretType}.${secret.secretName}`
```
galxy25, line 111: make this value `ae41d9cc-cd5c-4a90-8efe-0046784e195c` a constant so it can be reused by other secret methods (such as list or describe secret)

galxy25 (approved):

lib/identity/client.js
```diff
@@ -67,6 +68,60 @@ class Client extends PartialClient {
...
+    const re = new RegExp('^[a-zA-Z0-9-_]{1,30}$')
+    if (!re.test(trimmedName)) {
+      throw {
+        message: 'Secret name must contain alphanumeric characters, -, or _',
+      }
+    }
+    if (trimmedValue === '') {
+      throw { message: 'Value cannot be empty' }
+    }
+    const groupName = [`tozny.secret.${this.storage.config.clientId}`]
```
galxy25, line 87: This could also be a single string, and then on line 91 you can do [groupName] and then on line 97 not have to index into this array.

More importantly though, for group name lets have both the realm name and the secret type in the group name

`tozny.secret.{realmName}.{clientId}.{secretType}`

galxy25 (approved):
Approved pending tests and additional namespacing of group name

lib/identity/client.js
```diff
@@ -67,6 +68,60 @@ class Client extends PartialClient {
     options.headers.Authorization = `Bearer ${token}`
     return fetch(url, options)
   }
+
+  async createSecret(secret) {
+    const trimmedName = secret.secretName.trim()
+    const trimmedValue = secret.secretValue.trim()
+    if (trimmedName === '') {
+      throw { message: 'Name cannot be empty' }
```
lkwdwrd, line 76: if you are going to throw, make this an actual `Error`

```
throw new Error("Name cannot be empty")
```

lib/identity/client.js
lkwdwrd: I believe there is a limit of some kind on record-type names -- is there any chance with a namespace like this we'd start hitting it? I don't know what is is off the top of my head, but iirc, it is not that large.

lib/identity/client.js
```diff
@@ -67,6 +68,60 @@ class Client extends PartialClient {
...
+    }
+    const re = new RegExp('^[a-zA-Z0-9-_]{1,30}$')
```
lkwdwrd, line 78: you could just use a regex literal here, since it appears you only use it once.

```js
!/^[a-zA-Z0-9-_]{1,30}$/.test(trimmedName)
```

lib/identity/client.js
lkwdwrd: (This applies to the rest of the errors getting thrown as well)

lib/identity/client.js
```diff
@@ -67,6 +68,60 @@ class Client extends PartialClient {
...
+    let groupMembers = []
+    if (groups.groups.length < 1) {
+      // create the group if it doesn't exist
+      let newGroup = await this.storage.createGroup(groupName[0])
+      group = newGroup.group
+      let capabilities = {
+        read: true,
+        share: true,
+      }
+      var member = new GroupMember(this.storage.config.clientId, capabilities)
```
lkwdwrd, line 103: nit: either `let` or `const` is preferable for variable scoping. I generally use `const` first unless I need mutation, but I would just switch this to `let` to match the rest of the file.

lib/identity/client.js
lkwdwrd: 🤦 - oh, this is group name. Still, these keys are getting long, let's make sure we are working within our own limits for both group name and record type.

lib/identity/client.js
lkwdwrd: Is this a UUIDv5? Where ever it is defined, I would like a code comment or some form of documentation explaining how it works (why this UUID).

lib/identity/client.js
```diff
@@ -67,6 +68,60 @@ class Client extends PartialClient {
...
+      await this.storage.addGroupMembers(group.groupID, groupMembers)
+    } else {
+      group = groups.groups[0]
+    }
+    // create the record
+    const recordType = `tozny.secret.ae41d9cc-cd5c-4a90-8efe-0046784e195c.${secret.secretType}.${secret.secretName}`
+    const data = {
+      secretValue: secret.secretValue,
+    }
+    const meta = {
```
lkwdwrd, line 115: I see now we've switched back to `const` to declare variables. Like a lot of things, I care more about consistency than specific approach. My general rule of thumb for variable keywords goes like this:

- Use `const` unless you need the value to change
- Use `let` if you need the value to change
- Use `var` only as a last resort and document why the more precise scoping of let/const doesn't work in a comment

`let`/`var` will scope to the enclosing block (`{...}`) while `var` is function scoped (which can do weird things if you are not careful).

It's a messy world for variable keywords in JS. Even more messy because when `const` is an object (or array), _the object itself_ is mutable... `const` only guarantees the reference stays constant, not that the internal data is constant. 😅

lkwdwrd (approved):

lib/identity/client.js
carrala: this was just a randomly generated uuid, there's no special significance to this one; the goal was just to make sure that a client won't accidentally make another group with the exact same name. 
i can definitely add a comment about that if that's helpful.

galxy25: it can't be just "a" randomly generated uuid. It needs to be THE random uuid that is used for secrets across the js-sdk or any other SDK. Similar pattern that we used for note signatures

https://github.com/tozny/internal-docs/blob/ae0831ffc8bb45cca228114157dba86ba2a3faa7/tozny-platform/notes/tozny-field-signing.md#tozny-field-signing-protocol-with-blake2b-and-ed25519

carrala: okay, good to know. i will update that.

galxy25 (approved):

rosasck (approved):

lib/identity/client.js
efabens: @lkwdwrd do we need it to be a UUIDv5, can it just be a UUIDv4 that we know is constant across all of our SDKs that use secrets? At the end of the day we need to generate a UUIDv4 for the UUIDv5 so I am not sure what the additional value is. But this mostly just a curiosity of my own.

lkwdwrd: Down the line if we have a whole bunch of these namespaces, we can ship a single UUID namespace and calculate the UUIDv5 for each type within that namespace. It can be a little cleaner, and more meaningful, but it doesn't have to be. We already have a tozny UUID namespace from TFSV1, so it wouldn't be too hard to re-use that and just calculate the v5 for some string.

Ultimately, it doesn't matter one bit as long as they are well documented constants.

I don't feel strongly that it needs to be a v5, its just an option.

efabens: So you are thinking pseudocode wise TSV1 does
`namespace = "794253a4-310b-449d-9d8d-4575e8923f40"
sigtag = uuid.v5("TFSP1;ED25519;BLAKE2B", namespace)`

so toznysecrets should do
`namespace = "794253a4-310b-449d-9d8d-4575e8923f40"
sigtag = uuid.v5("tozny.secrets", namespace)`

I think that makes good sense.
Any reason an online tool like https://www.uuidtools.com/v5 can't be used to generate it? it doesn't seem the standard golang implementation supports it.

@lkwdwrd

lkwdwrd: exactly -- and yes that online generator is just fine to create it. We can even hard code the UUID for the time being with a code comment to indicate where it came from. Calculation becomes more valuable as we have more of them, but I have a sneaking suspicion as we go forward... we will have more of these.

lkwdwrd (approved):

lkwdwrd (approved):
Re-reviewed -- this is also looking really good. Nothing further from me here for the code 👏 

The only thing I noticed that needs updating prior to merge -- bring back the package-lock.json. It is important we keep that version controlled.

__tests__/secret.test.js
```diff
@@ -0,0 +1,180 @@
...
+  })
+  it('can create a secret, and list it', async () => {
+    const secret = {
+      secretType: 'Credential',
+      secretName: `test-secret-${uuidv4()}`,
+      secretValue: 'secret-value',
+      description: 'this is a description',
+    }
+    await ops.createSecret(realmConfig, identity, secret)
+    await new Promise(r => setTimeout(r, 1000))
```
lkwdwrd, line 107: In an older PR I posted an example of a helper function which will keep these running faster and retry a number of times. I'd like to see us switch to something like that instead of just a 1 second timeout. I'm wondering if this may be part of the reason our tests are failing in some environments. 🤔 

ref: https://github.com/tozny/js-sdk/pull/41/files#r572278849

```js
/**
 * Get the results from a query, using a loop to ensure they have time to index
 */
async function waitForNext(query) {
  // Start with a very short delay as immediate fetch of results right after writing
  // sometimes failes, but even with a very short window of wait it can succeed
  // first try.
  await new Promise(r => setTimeout(r, i * 100))
  // 10-second timeout period, 20 rounds of half a second between tries.
  for (let i = 0; i < 20; i++) {
    const found = await query.next()
    if (found.length > 0) {
      break
    }
    // delay half a second between tries
    await new Promise(r => setTimeout(r, i * 500))
  }
  return found
}
```

rosasck:
Added Luke's function!

rosasck:
ran run build, hoping that fixes the travis. If not ill have to dig

rosasck:
Realized that Search Service order by is not deployed... i will deploy on monday! and get this build not failing!

__tests__/utils/operations.js
```diff
@@ -738,19 +738,23 @@ module.exports = {
...
+          })
       },
       JSON.stringify(config),
       user.stringify(),
       secretName,
       secretType,
       usernameToAdd
     )
-    return result
+    return JSON.parse(result)
```
lkwdwrd, line 757: Is this just a plain object? Should it parse to a type at all?

rosasck: its a plain object with a recordType, encrypted access key and public key and

lkwdwrd: I've been thinking we may want to create actual secret types at some point, but we can revisit this then.

PR #41, opened 2021-02-06
List tozsecrets

rosasck:
This is based off create-secrets, not master


lkwdwrd:
Wish I knew why your computer was so adamant about getting rid of `package-lock.json` -- we need that file 😅

lib/identity/client.js
```diff
@@ -126,6 +126,24 @@ class Client extends PartialClient {
...
+    const request = new Search(true, true, limit)
+    let searchType = `tozny.secrets.${SECRET_UUID}.*`
+    request.match({ type: searchType }, 'OR', 'FUZZY')
+    const resultQuery = await this.storage.search(request)
+    const secretList = await resultQuery.next()
+    console.log(secretList)
+    return secretList, resultQuery.request.nextToken
+  }
+  async loadMoreSecrets(limit, nextToken) {
+    const request = new Search(true, true, limit)
```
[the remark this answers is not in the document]
galxy25: I'm not opposed to that.

galxy25 (approved):

rosasck:
Just Saw @lkwdwrd comment! just fixed it and will fix the js-sdk to work with the result query

__tests__/secret.test.js
```diff
@@ -89,4 +89,15 @@ describe('Tozny identity client', () => {
...
   })
+  it('can create a secret, and list it', async () => {
+    const secret = {
+      secretType: 'Credential',
+      secretName: `test-secret-${uuidv4()}`,
+      secretValue: 'secret-value',
+      description: 'this is a description',
+    }
+    await ops.createSecret(realmConfig, identity, secret)
+    await new Promise(r => setTimeout(r, 1000))
```
lkwdwrd, line 100: hmm -- I admit I haven't paid as much attention on tests and should have. Why do we need this Timeout? I think in the search we have a method to have searches try until they find their results for a set amount of time or something (guessing it is here to allow search service to populate). This is a little less brittle than a timeout, which is not guaranteed by search.

__tests__/secret.test.js
```diff
@@ -89,4 +89,15 @@ describe('Tozny identity client', () => {
...
+    await ops.getSecrets(realmConfig, identity, 10)
```
lkwdwrd, line 101: Do we validate our expected secrets come back in the search?

lkwdwrd (approved):
May need some slight test adjustments, but code is 👍

__tests__/secret.test.js
rosasck: This was the solution @willmichael  and i could find, since a secret is added and immediately is searched. We needed the small gap to allow search to propagate

__tests__/secret.test.js
rosasck: It returns a query so its hard to write a "expects" but yes it does come back

__tests__/secret.test.js
lkwdwrd: https://github.com/tozny/js-sdk/blob/f0254f256f8dc996c0adb4d3e660df4d1ee1d8b0/__tests__/search.test.js#L28 - is where the loop is. This could get refactored into a search utility which runs the next in a loop until the results come back or it times out, and fail only if the timeout is hit.

It's basically an improvement on the straight timeout by allowing you to set more potential time for indexing to occur while still allowing things to move more quickly when indexing occurs quickly.

e.g.
```js
/**
 * Get the results from a query, using a loop to ensure they have time to index
 */
async function waitForNext(query) {
  // Start with a very short delay as immediate fetch of results right after writing
  // sometimes failes, but even with a very short window of wait it can succeed
  // first try.
  await new Promise(r => setTimeout(r, i * 100))
  // 10-second timeout period, 20 rounds of half a second between tries.
  for (let i = 0; i < 20; i++) {
    const found = await query.next()
    if (found.length > 0) {
      break
    }
    // delay half a second between tries
    await new Promise(r => setTimeout(r, i * 500))
  }
  return found
}
  ```

PR #42, opened 2021-02-09
Update secrets

rosasck:
Update secret JS-SDK support plus test! 
- I couldn't find a better way for the test to prove it updated it, so its a bit sloppy


__tests__/secret.test.js
```diff
@@ -110,4 +110,26 @@ describe('Tozny identity client', () => {
...
+    await new Promise(r => setTimeout(r, 1000))
+    const query = await ops.getSecrets(realmConfig, identity, 10)
+    let secrets = await query.next()
+    await ops.updateSecret(realmConfig, identity, secret)
+    await new Promise(r => setTimeout(r, 10000))
+    const query2 = await ops.getSecrets(realmConfig, identity, 100)
+    const newSecrets = await query2.next()
+    let length1 = secrets.length
+    let length2 = newSecrets.length
+    expect(length1 + 1).toBe(length2) // this shows we created a new secret
```
galxy25, line 130: Nit: the value you are testing should be on the left side of this expect, as it is for all the expects below

```js
 expect(length2).toBe(length1 + 1)
```

__tests__/secret.test.js
```diff
@@ -110,4 +110,26 @@ describe('Tozny identity client', () => {
...
+    expect(secrets[length1 - 1].meta.plain.version).toBe('latest') // this shows that the last item is latest
```
galxy25, line 131: Your comment doesn’t seem to line up with the code here, secrets is the original list of secrets after you created the first one (think of you how you can rename these variables to be more meaningful and clear

 Secrets could be `singleSecretVersionListResponse`
NewSecrets could be `multipleSecretVersionsListResponse`

)

lib/identity/client.js
```diff
@@ -141,6 +139,35 @@ class Client extends PartialClient {
...
+    const secretList = await resultQuery.next()
+    // Find the Record Tagged Latest
+    let recordId
+    for (var i = 0; i < secretList.length; i++) {
+      if (secretList[i].meta.plain['version'] == 'latest') {
+        recordId = secretList[i].meta.recordId
+      }
+    }
+    // Update latest Record to be new version number
+    const versionNumber = resultQuery.totalResults
```
galxy25, line 157: This is assuming too much, what if someone has deleted one of the secret versions, then the count of results will not equal the version. Instead you should get the version number based on what the version number is for the secret/record tagged “latest”

__tests__/secret.test.js
```diff
@@ -110,4 +110,26 @@ describe('Tozny identity client', () => {
...
+    expect(newSecrets[length1 - 1].meta.plain.version).toBe('v1') // this shows that now that item changed to V1
+    expect(newSecrets[length2 - 1].meta.plain.version).toBe('latest') // and the newest item is now latest
```
galxy25, line 133: Alright, this makes it clear I told you the wrong thing for how to version secrets, so lets try version 2 of secrets versioning

Every secret has a tag “version”, and a tag “version_number”. 
When the secret is first created, version = latest, version_number=1

When that secret is updated
Old secret version is updated to be archived, version number is unchanged
New secret is tagged with version=latest, version_number=2

lib/identity/client.js
```diff
@@ -141,6 +139,35 @@ class Client extends PartialClient {
     const resultQuery = await this.storage.search(request)
     return resultQuery
   }
+  async updateSecret(secret) {
+    // Search for all records of the same type
+    const request = new Search(true)
+    let searchType = `tozny.secret.${SECRET_UUID}.${secret.secretType}.${secret.secretName}`
+    request.match({ type: searchType })
+    const resultQuery = await this.storage.search(request)
```
galxy25, line 147: 1.) you shouldn’t have to do a search to find the previous latest version of this secret because the user is passing in the secret to update
2.) If you do need to find that secret from a search, talk to @willmichael about how to construct this search to be more focused (e.g. searching for records not only of the record type but with a tag of version:latest)

lib/identity/client.js
```diff
@@ -141,6 +139,35 @@ class Client extends PartialClient {
...
+    }
+    // Update latest Record to be new version number
+    const versionNumber = resultQuery.totalResults
+    const updatedMetaVersion = {
+      secretType: secret.secretType,
+      secretName: secret.secretName,
+      description: secret.description,
+      version: `v${versionNumber}`,
+    }
+    const read = await this.storage.readRecord(recordId)
```
galxy25, line 164: Read, read what? This is not a good variable name.

lib/identity/client.js
```diff
@@ -141,6 +139,35 @@ class Client extends PartialClient {
     const resultQuery = await this.storage.search(request)
     return resultQuery
   }
+  async updateSecret(secret) {
```
galxy25, line 142: This should take in the old secret to update, and the new value to update it with

galxy25, line 142: Add function documentation please

galxy25 (changes requested):
See comments

__tests__/secret.test.js
efabens: I think having a true latest tag that moves adds a lot of state and complexity that we don't want to manage, records also have version uuids so if someone were to be verifying a record based on this that would become out of date (I don't know if anyone would do that but it is something to think about). The original python version of secrets does use numbers for versions, but we have dates for when records are created, I don't really see why a counting versioning system is better than just a date based one.

lib/identity/client.js
```diff
@@ -141,6 +139,35 @@ class Client extends PartialClient {
...
+    const versionNumber = resultQuery.totalResults
+    const updatedMetaVersion = {
+      secretType: secret.secretType,
+      secretName: secret.secretName,
+      description: secret.description,
+      version: `v${versionNumber}`,
+    }
+    const read = await this.storage.readRecord(recordId)
+    read.meta.plain = updatedMetaVersion
+    await this.storage.updateRecord(read)
```
efabens, line 166: I don't think we have a need to or should be updating records. When "updating secrets" It is really just creating a new record of the same type and record name.

__tests__/secret.test.js
galxy25: I may be trying to pre-maturely optimize something that isn't a problem, but I want a clear and fast way to say "get me the latest version of the secret" without having to fetch all the records, sort them by date, and take the newest one. So as long as we can keep the interface that developers or users want - a single method to get a secret that always returns the latest version, I'm fine with the date based approach or the "latest" tag way I originally proposed.

efabens: 100% agree on needing a simple way to ask that question. I would argue record type + write date limit 1 gets you there

galxy25: @rosasck talk with Eli about this and then implement the approach you two agree on please

rosasck:
Updated updateSecrets!

efabens (approved):

rosasck:
sad path tests commint

rosasck:
Sad path tests added

lib/identity/client.js
```diff
@@ -123,24 +128,52 @@ class Client extends PartialClient {
     const data = {
       secretValue: secret.secretValue,
     }
+    let timestamp = Date.now().toString()
     const meta = {
       secretType: secret.secretType,
       secretName: secret.secretName,
       description: secret.description,
-      version: 'latest',
+      version: timestamp,
```
willmichael, line 136: @efabens I assume this is what you guys settled instead of version: 'latest'; however, I don't think this provides much value over our created at timestamp since plain_meta is string based and not order-able

efabens: Yeah doesn't provide value for elastic search ordering but provides value as a piece of data that is constant that a consumer could use. It was between using timestamp or not having a value

galxy25: @efabens we do want to be able to show users both in the cli and the ui a version NUMBER, I'm fine if we use timestamps under the hood but keep that constraint in mind

efabens: @galxy25 given that we can delete records I don't think that we can guarantee that an iterating number can be trusted to always be correct. If in a user interacting situation we want an easy way to list out version numbers than I think the UI should order and create temporary values that match versions. For an automated system where they know a particular version is wanted than the timestamp provides a constant that can be trusted

PR #43, opened 2021-02-09
Get latest secret

rosasck:
Gets the latest version of a secret or empty if the secret doesn't exist
tests included


__tests__/secret.test.js
```diff
@@ -177,4 +177,37 @@ describe('Tozny identity client', () => {
...
+    await new Promise(r => setTimeout(r, 10000))
+    let latestVersion = await ops.getLatestSecret(
+      realmConfig,
+      identity,
+      `test-secret-updatetest35558800`,
+      'Credential'
+    )
+    expect(latestVersion.data.secretValue).toBe('updatedSecretValue')
+  })
+  it('get the latest version of an invalid secret', async () => {
```
galxy25, line 204: `gets`

__tests__/secret.test.js
```diff
@@ -177,4 +177,37 @@ describe('Tozny identity client', () => {
       ops.updateSecret(realmConfig, identity, oldSecret, newSecret)
     ).rejects.toThrow()
   })
+  it('get the latest version of a secret', async () => {
```
galxy25, line 180: `gets`

lib/identity/client.js
```diff
@@ -155,6 +155,27 @@ class Client extends PartialClient {
...
+    const request = new Search(true)
+    let searchType = `tozny.secret.${SECRET_UUID}.${secretType}.${secretName}`
+    request.match({ type: searchType })
+    let resultQuery = await this.storage.search(request)
+    let resultList = await resultQuery.next()
+    let lengthOfResults = resultList.length
+    if (resultList.length != 0) {
+      return resultList[lengthOfResults - 1]
+    }
+    return resultList[0]
```
galxy25, line 176: so I guess the caller of this code just always has to check this response is not null? Can we return a struct that has a field for if the secret exists at all, and a field for holding that secret (if it does exist)

lib/identity/client.js
```diff
@@ -155,6 +155,27 @@ class Client extends PartialClient {
...
+   * @param {string} secretType   The type of the Secret chosen by User.
+   *
+   *
+   * @return {Promise<Record>}
+   */
+  async getLatestSecret(secretName, secretType) {
+    const request = new Search(true)
+    let searchType = `tozny.secret.${SECRET_UUID}.${secretType}.${secretName}`
+    request.match({ type: searchType })
+    let resultQuery = await this.storage.search(request)
```
galxy25, line 170: Since search is sorting these ascending, wouldn't the first secret it returned be the oldest version (not the newest)?

galxy25: https://github.com/tozny/e3dbSearchService/blob/master/clients/elastic_query.go#L134
https://github.com/tozny/api-specs/blob/master/search/swagger.yaml#L241

rosasck: Yeah! my work around was to take the last record returned

galxy25: Prefer updating search as discussed https://tozny.slack.com/archives/G0DPBU8T1/p1612902038062200 to support this use case (we will definitely re-use & benefit from that search capability in other areas)

You're current code is making the assumption that there isn't more than a few hundred versions of the secret, instead of checking to see if it needs to page to get more search results, but if you update search you won't have to worry about that ;-)

lib/identity/client.js
```diff
@@ -155,6 +155,27 @@ class Client extends PartialClient {
...
+    let resultList = await resultQuery.next()
+    let lengthOfResults = resultList.length
+    if (resultList.length != 0) {
+      return { exists: true, result: resultList[lengthOfResults - 1] }
+    }
+    return { exists: false, result: resultList[0] }
```
galxy25, line 176: can result be null instead / won't this blow up if the list is empty and you try and index the first item (which won't exist because its empty...)

lib/identity/client.js
rosasck: i read that message wrong the first time, no wonder trying to get descending to work, did not

types/searchRange.js
```diff
@@ -1,12 +1,13 @@
...
 
 const keyWhitelist = ['CREATED', 'MODIFIED']
+const sortOrderWhitelist = ['ASCENDING', 'DESCENDING']
 
 /**
  * Create a search query against Tozny Platform.
  */
 class SearchRange extends Serializable {
-  constructor(start, end, key = 'CREATED') {
+  constructor(start, end, sort_order = 'ASCENDING', key = 'CREATED') {
```
efabens, line 10: The new parameter should be placed at the end so that existing implementations don't get unexpected behavior

lib/identity/client.js
```diff
@@ -166,6 +166,27 @@ class Client extends PartialClient {
...
+   * @param {string} secretName   The name of the Secret given by User.
+   * @param {string} secretType   The type of the Secret chosen by User.
+   *
+   *
+   * @return {Promise<Record>}
+   */
+  async getLatestSecret(secretName, secretType) {
+    const request = new Search(true)
+    let searchType = `tozny.secret.${SECRET_UUID}.${secretType}.${secretName}`
+    request.match({ type: searchType }, 'AND', 'EXACT')
```
efabens, line 180: Should this now have a range that is descending?

lib/identity/client.js
```diff
@@ -166,6 +166,27 @@ class Client extends PartialClient {
...
+    let resultQuery = await this.storage.search(request)
+    let resultList = await resultQuery.next()
+    console.log(resultList)
+    if (resultList.length > 0) {
+      resultList[resultList.length - 1]
```
efabens, line 185: I think we want to be able to grab item 0 here such that we know even if there are multiple "pages" of results we have the newest

lib/identity/client.js
```diff
@@ -166,6 +166,27 @@ class Client extends PartialClient {
...
+   *
+   * @return {Promise<Record>}
+   */
+  async getLatestSecret(secretName, secretType) {
+    const request = new Search(true)
+    let searchType = `tozny.secret.${SECRET_UUID}.${secretType}.${secretName}`
+    request.match({ type: searchType }, 'AND', 'EXACT')
+    let resultQuery = await this.storage.search(request)
+    let resultList = await resultQuery.next()
+    console.log(resultList)
```
efabens, line 183: errant console log that should get removed before merging

rosasck:
Reworked for changes with Luke and now it works and uses the search end point changes

types/searchRange.js
```diff
@@ -17,6 +17,7 @@ class SearchRange extends Serializable {
     if (!(end instanceof Date)) {
       throw new Error('end must be an instance of Date')
     }
+
```
efabens, line 20: Doesn't this still need the new constructor parameter?

__tests__/secret.test.js
```diff
@@ -193,6 +193,53 @@ describe('Tozny identity client', () => {
...
+      }
+      // delay 200 milliseconds between tries
+      await new Promise(r => setTimeout(r, 200))
+    }
+    expect(latestVersionOfSecret.exists).toBe(true)
+    expect(latestVersionOfSecret.results.data.secretValue).toBe(
+      'updatedSecretValue'
+    )
+  })
+  it('gets the latest version of an invalid secret', async () => {
```
galxy25, line 234: should this be `it doesn't`?

lib/identity/client.js
```diff
@@ -166,6 +166,26 @@ class Client extends PartialClient {
     return resultQuery
   }
   /**
+   * getLatestSecrets
```
galxy25, line 169: `getLatestSecret`

galxy25 (approved):

types/searchRange.js
rosasck: We took it out of SearchRange and put it in SearchOrder

__tests__/secret.test.js
```diff
@@ -193,6 +193,53 @@ describe('Tozny identity client', () => {
...
+      secretType: 'Credential',
+      secretName: testName,
+      secretValue: 'updatedSecretValue',
+      description: 'this is a description',
+    }
+    await ops.createSecret(realmConfig, identity, oldSecret)
+    await ops.updateSecret(realmConfig, identity, oldSecret, newSecret)
+    const start = new Date()
+    let latestVersionOfSecret
+    while (new Date() - start < 30000) {
```
lkwdwrd, line 214: I see now why the waitForNext method would not work here... The functional programmer in me says there is a pattern we can abstract out to be even more general, but let's hold off until we see another area that is needed :)

Good job adapting the retry loop for your needs!

__tests__/utils/operations.js
```diff
@@ -746,4 +746,35 @@ module.exports = {
...
+            return secret
+          })
+      },
+      JSON.stringify(config),
+      user.stringify(),
+      secretName,
+      secretType
+    )
+    if (secret.exists == true) {
+      let secretResult = await Tozny.types.Record.decode(secret.results)
```
lkwdwrd, line 775: 👍 - yay, proper types after transport!

__tests__/secret.test.js
galxy25: Lol I had to resist the urge to @ you on this and ask for how to do this with currying, I know it's possible in js

types/search.js
```diff
@@ -32,8 +33,12 @@ class Search extends Serializable {
     return this
   }
 
-  range(start, end, key) {
-    this.searchRange = new SearchRange(start, end, key)
+  range(start, end, key, sort_order) {
```
lkwdwrd, line 36: I think this chunk for range can get reverted with the `searchOrder` type, yes?

lkwdwrd (approved):
Looking good. Some minor cleanup, but mostly kudos!

efabens (approved):

PR #45, opened 2021-02-10
Adjustments for stable tests

lkwdwrd:
I believe this set of changes will stabilize our test runs from Travis. It seems like a lot, but the changes are all pretty small tweaks, really minor. Biggest thing is there was some instability in going to and from the test environment and waiting for search results to get indexed. (Also found some bugs in our test configuration code)

Updates:

- Make sure local test options are respected as boolean values
- Update wait for next helper
  - switch to a while loop for more precise timeout (turned out it wasn't needed, but I kept it more precise)
  - short circuit the request if it is already recorded as done
  - make sure to reset pagination on each loop
  - take a test function which defaults to the previous length check to allow use-case specific testing of expected search results
- Define secrets specific helper which waits for secrets with a specific name to index, and filters search results to only contain secrets with that name
- Update names to use a local variable in each secret test
- Adjust the ops getSecrets() to run the methods in the test environment only, including running next
- Switch tests not directly testing getSecrets to use the direct identity client `getSecrets()` method
- Consume new wait for next helper in search tests

### DIFF
--- a/__tests__/secret.test.js
+++ b/__tests__/secret.test.js
@@ -37,9 +37,10 @@ beforeAll(async () => {
 
 describe('Tozny identity client', () => {
   it('can create a secret', async () => {
+    const testName = `test-secret-${uuidv4()}`
     const secret = {
       secretType: 'Credential',
-      secretName: `test-secret-${uuidv4()}`,
+      secretName: testName,
       secretValue: 'secret-value',
       description: 'this is a description',
     }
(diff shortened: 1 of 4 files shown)


__tests__/secret.test.js
```diff
@@ -169,3 +179,32 @@ describe('Tozny identity client', () => {
...
+ * secret are found. By default it ensures at least one secret of the given name
+ * exists before returning.
+ *
+ * If the result gets to the end of the 10 second timeout period, this method throws.
+ *
+ * @param {SearchResult} query A secrets search result
+ * @param {string} name The name of the secrets being looked for
+ * @param {int} numRequired The number of secrets which should be found before returning
+ */
+async function waitForNextOfName(query, name, numRequired = 1) {
```
galxy25, line 196: sweet

galxy25 (approved):

efabens (approved):

rosasck (approved):

rosasck:
Thank you so much Luke!!!

PR #46, opened 2021-02-13
Share Secrets by identity username

rosasck:
Tests and JS SDK support for sharing a secret Type with username


lib/identity/client.js
```diff
@@ -205,6 +206,62 @@ class Client extends PartialClient {
...
+    const groupList = await this.storage.listGroups(
+      this.storage.config.clientId,
+      [groupName]
+    )
+    if (groupList.groups.length < 1) {
+      throw new Error('Group Does Not Exist')
+    }
+    const currentGroup = groupList.groups[0]
+    // Create the request for username to client id map
+    const SearchRealmIdentitiesRequest = {
```
efabens, line 232: Doesn't this already have an sdk method? If not it should be extracted out.

lib/identity/client.js
```diff
@@ -205,6 +206,62 @@ class Client extends PartialClient {
...
+   *
+   * @param {String} secretType
+   * @param {String} secretName
+   * @param {Array} usernameToAdd
+   *
+   *
+   */
+  async shareSecretWithUsername(secretName, secretType, usernamesToAdd = []) {
+    // Generate Group to Validate it exists
+    const groupName = `tozny.secret.${this.config.realmName}.${
```
efabens, line 220: I am not convinced that every single secret type needs its own group. Each user has a group for all their secrets and then secrets can be added to other groups. But it seems that for direct shares it makes more sense to just share with clients directly.

lib/identity/client.js
```diff
@@ -205,6 +206,62 @@ class Client extends PartialClient {
...
+    )
+    const clientMapping = await validateResponseAsJSON(response)
+    // Create an array and set capabilities to read
+    let groupMembers = []
+    const capabilities = {
+      read: true,
+    }
+    usernamesToAdd.forEach(username => {
+      let usernameLowerCase = username.toLowerCase()
+      let clientID =
```
efabens, line 255: this has a chance of failing if a username provided does not have an associated client. You should verify that it exists.

__tests__/secret.test.js
```diff
@@ -241,6 +242,39 @@ describe('Tozny identity client', () => {
     expect(latestVersion.exists).toBe(false)
   })
 })
+it('can create a secret and share it with a username', async () => {
```
efabens, line 245: let's have another test for sharing with a username that doesn't exist and that it fails smoothly

lib/identity/client.js
rosasck: I was following what is written in create Secret, but i can alter that method if we just want one group per person?

lib/identity/client.js
rosasck: nope! just in the identity portal! but ill make it now!

willmichael (approved):
ditto on the sad path testing from eli

rosasck:
This is being changed!

rosasck:
Updated to make a new group with the client they are sharing with! since i wanted only one group for each paring i didnt wanna use the username of the client being shared with since then, when the new client wants to share with this client that would result in a second group! 
the group name is `tozny.secret.${this.config.realmName}.${this.storage.config.clientId}.${clientID}`

rosasck:
Also added a sad path test

galxy25:
> Updated to make a new group with the client they are sharing with! since i wanted only one group for each paring i didnt want to use the username of the client being shared with since then, when the new client wants to share with this client that would result in a second group!
> the group name is `tozny.secret.${this.config.realmName}.${this.storage.config.clientId}.${clientID}`

we (I) do want a different group for each pairing, and I especially don't want any ambiguity in the system, because that will eventually become a security issues. With this group naming, how do I know which secrets are shared from the client id on the left to the client id on the right and vice versa? In reality, doesn't matter because for right now we're only allowing the client the secret is shared with to read, not write new records into this group.

rosasck:
Added Username to the group name to be able to distinguish that this group is things shared with that user

galxy25:
> Added Username to the group name to be able to distinguish that this group is things shared with that user

we actually do want client id vs username, as username is NOT unique across realms, but client id will be. Remember to keep your consumer in mind, for the UI we want to display human readable things like usernames and group names, but from the api and inner working of the sdk we want to deal in the individual resource ids (group or client id)

galxy25 (approved):

PR #47, opened 2021-02-17
Allow passing of options when requsting a password reset

lkwdwrd:
Also switches back the default template to reset_password... options will work to send the claim_account template.


efabens (approved):

lib/identity/realm.js
```diff
@@ -567,15 +567,21 @@ class Realm extends CryptoConsumer {
    * A wrapper around the broker login flow used for email account recovery.
    *
    * @param {string} username The username to recover.
+   * @param {object} options The options for to send when issuing the challenge to recover the account
```
willmichael, line 570: The options to send

willmichael (approved):

PR #48, opened 2021-02-18
adding storage endpoint for unsharing record with group

rosasck:
This includes the support for the storage endpoint! secret support coming soon!


lib/storage/client.js
```diff
@@ -1465,5 +1465,38 @@ class Client extends CryptoConsumer {
...
+      }
+    )
+    // Check the response code
+    switch (response.status) {
+      case 204:
+        return true
+      case 403:
+        throw new Error('Unauthorized')
+      default:
+        throw new Error('Error while unsharing record with group')
```
willmichael, line 1498: We should try to pass up the status code of the error for visibility

lib/storage/client.js
```diff
@@ -1465,5 +1465,38 @@ class Client extends CryptoConsumer {
...
+
+  /**
+   * removeSharedRecordWithGroup
+   *
+   * @param {string} groupId   UUID assigned by TozStore, used to identify a group.
+   * @param {string} recordType Record type to be unshared with the group.
+   *
+   * @return {Promise<Record>}
+   */
+  async removeSharedRecordWithGroup(groupId, recordType) {
```
willmichael, line 1477: potentially better to have this be "revokeRecordWithGroup" because it'll match our other sdk methods (share/revoke)

willmichael (approved):

__tests__/secret.test.js
```diff
@@ -290,6 +290,96 @@ describe('Tozny identity client', () => {
...
+    }
+    const list = await ops.getSecretSharedList(
+      realmConfig,
+      identity,
+      testName,
+      'Credential'
+    )
+    expect(list[0].username).toBe(testUsername)
+  })
+  it('can return an empty list if not shared', async () => {
```
efabens, line 366: can you update this test to share and then unshare and show that it doesn't get returned, or create a new test that does that.

lib/identity/client.js
```diff
@@ -300,6 +306,95 @@ class Client extends PartialClient {
...
+    // Look up group for our current user sharing to the username
+    // The group name is ordered by first clientID is the client who wrote the secret
+    // and second clientID is the user that we are sharing with
+    // This allows visibility to who owns the secret within a pairing
+    const groupName = `tozny.secret.${this.config.realmName}.${this.storage.config.clientId}.${clientID}.${secretName}.${secretType}`
+    const groupList = await this.storage.listGroups(
+      this.storage.config.clientId,
+      [groupName]
+    )
+    if (groupList.groups.length < 2) {
```
efabens, line 343: why isn't this 1? I assume I am missing something.

lib/identity/client.js
```diff
@@ -300,6 +306,95 @@ class Client extends PartialClient {
...
+ 
+   *
+   * @return {Promise<Record>}
+   */
+  async getSecretSharedList(secretName, secretType) {
+    const groupList = await this.storage.listGroups(
+      this.storage.config.clientId
+    )
+    if (groupList.groups.length < 1) {
+      // If there is only one group, this means its the automatic group, no shared groups
```
lkwdwrd, line 372: I'm not really understanding the 'automatic' group?

lib/identity/client.js
```diff
@@ -300,6 +306,95 @@ class Client extends PartialClient {
...
+      return []
+    } else {
+      let groupSharedList = []
```
lkwdwrd, line 375: Not explicitly necessary, but I think this would be easier to follow as a map instead of a loop, and end up less code.

lib/storage/client.js
```diff
@@ -1465,5 +1465,41 @@ class Client extends CryptoConsumer {
...
+        },
+        body: JSON.stringify(removeRecordSharedWithGroupRequest),
+      }
+    )
+    // Check the response code
+
+    switch (response.status) {
+      case 204:
+        return true
+      case 403:
```
lkwdwrd, line 1496: probably 401

lib/identity/client.js
```diff
@@ -300,6 +306,95 @@ class Client extends PartialClient {
...
+      for (let index = 0; index < groupList.groups.length; index++) {
+        let groupNameSplit = groupList.groups[index].groupName.split('.')
+        if (groupNameSplit.length == 7) {
```
lkwdwrd, line 378: nit: maybe a `!=` with a `continue` statement? Will keep the code less nested which can make it easier to read.

Also: If doing a `.map()`, this would become a `.filter()` to remove items which don't match the criteria.

```js
.map(
  g =>  { /*do any needed transformation, maybe return false if it doesn't match the criteria*/ }
)
.filter( g => g ) //identity function will filter out falsey values, removing the items marked as false
```

lib/identity/client.js
```diff
@@ -300,6 +306,95 @@ class Client extends PartialClient {
...
+          if (
+            groupNameSplit[3] == this.storage.config.clientId &&
+            groupNameSplit[5] == secretName &&
+            groupNameSplit[6] == secretType &&
+            groupList.groups[index].memberCount > 1
+          ) {
+            // Check to make sure they are the writer
+            // and secret type and secret name are the same
+            groupSharedList.push({
+              username: groupList.groups[index].description,
```
lkwdwrd, line 389: description feels a little odd - do we not have a place to store structured metadata?

lkwdwrd (changes requested):
Some thoughts inline. Only blocking issue, really is the Array reference potential bug.

lib/identity/client.js
rosasck: When you create a secret a group for you is created to store all your secrets, so i wanted to have an easy escape case but if its harder to follow i can get rid of it

lib/identity/client.js
rosasck: unfortunently, not for groups

lib/identity/client.js
```diff
@@ -300,6 +306,98 @@ class Client extends PartialClient {
...
+      return revokedFromGroup
+    }
+  }
+
+  /**
+   * getSecretSharedList returns a list of usernames to which the secret is shared with 
+   *
+   * @param {String} secretType secret type that is being used to see the share list
+   * @param {String} secretName secret name that is being used to see the share list
+
```
lkwdwrd, line 364: nit: remove extra line

lib/storage/client.js
```diff
@@ -1196,6 +1196,42 @@ class Client extends CryptoConsumer {
...
     return returnValues
   }
+  /**
+   *
+   * @param {string} clientID   UUID assigned by TozStore, used to identify the client; optional.
+   * @param {String} groupName Group name you want to get information regarding.
+   *
+   * @return {Object}  The group object or null if it does not exist.
+   */
+  async groupInfo(clientID = null, groupName) {
```
lkwdwrd, line 1206: nit: reverse parameters so you can call `groupInfo('groupName') without the need to set clientID.

Also: default value of non-passed arguments is `undefined` so you can probably omit the null value and just check for a falsey value.

lib/storage/client.js
```diff
@@ -1196,6 +1196,42 @@ class Client extends CryptoConsumer {
...
+    const urlData = {
```
lkwdwrd, line 1207: probably a bit nicer to populate the object entirely before running the `urlEncodeData()` function -- it feels a bit odd to run it and then append strings. Also, it runs your values through the encoder to make sure they are url-safe strings for final output.

lib/storage/client.js
```diff
@@ -1196,6 +1196,42 @@ class Client extends CryptoConsumer {
...
+      {
+        method: 'GET',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      }
+    )
+    let groupsJson = await validateResponseAsJSON(response)
+    let finalgroup
+    groupsJson.groups.forEach(group => {
```
lkwdwrd, line 1225: `find()` will be cleaner than `foreach()` (plus is more efficient, though it's a total micro optimization that doesn't really matter much): 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

Might also consider checking to make sure the group was actually found and erroring on the not-found case.

```js
const finalGroup = groupsJson.groups.find(g => g.group_name == groupName)
if (!finalGroup) {
  // handle error
}
return {
  group: Group.decode(finalGroup)
}
```

lib/identity/client.js
lkwdwrd: hmm, something to add to the backlog I suppose

lib/storage/client.js
lkwdwrd: Also - what is the value of the object wrapping the returned group? I think in this instance it should just return the `Group` type directly unless I am missing something?

lkwdwrd (approved):
Some minor suggestions for additional cleanup, but looks good.

lib/identity/client.js
rosasck: totally agree!

lkwdwrd (approved):

galxy25 (approved):

PR #49, opened 2021-02-19
File secrets

carrala:
tests coming soon


rosasck (approved):
I learned what a kikibyte is

efabens (approved):

lib/identity/client.js
```diff
@@ -4,6 +4,7 @@ const fetch = require('isomorphic-fetch')
 const PartialClient = require('./partialClient')
 const { SECRET_UUID, TYPES } = require('../../lib/utils/constants')
 const { GroupMember, Search } = require('../../types')
+const { fileAsUrl } = require('../../browser/helpers')
```
lkwdwrd, line 7: we should not pull browser helpers into `/lib` code. 

All of the library code is shared between node and browser targets. All platform specific methods need to live in their respective platform directories.

If there is a generic helper, it can get added to the library helper file

lib/identity/client.js
```diff
@@ -212,6 +230,18 @@ class Client extends PartialClient {
...
     return newSecretCreated
   }
+  /**
+   * downloadFile
+   *
+   * @param {Object} recordId
+   *
+   * @return {string}
+   */
+  async downloadFile(recordId) {
```
lkwdwrd, line 240: yeah, instead of this, probably return the platform agnostic `File()` object and in documentation demonstrate the helpers for turning that into items for use for the demonstrated platform.

lkwdwrd (changes requested):

carrala:
oops, pushed preemptively! -- it's using a file that i have locally, but i'll change this so it will work with the travis tests!

efabens (approved):

lib/identity/client.js
```diff
@@ -233,14 +232,13 @@ class Client extends PartialClient {
...
-   * @param {Object} recordId
+   * @param {Object} record the record that contains the file info or the recordId
    *
    * @return {string}
    */
-  async downloadFile(recordId) {
-    const file = await this.storage.getFile(recordId)
-    const url = await fileAsUrl(file)
-    return url
+  async downloadFile(record) {
```
lkwdwrd, line 239: nit: Since this is a direct client method, I'm wondering if this method is descriptive enough. Perhaps we just use `getFile` so it's an alias to the storage method, just allowing easier access.

__tests__/utils/operations.js
```diff
@@ -814,4 +814,23 @@ module.exports = {
     }
     return secret
   },
+  async downloadFile(config, user, recordId) {
```
lkwdwrd, line 817: This is definitely not going to work in a browser context. We'll need to think more about how to best test the file features. I think there are ways, but we're gonna have to potentially do some environment detection and drop to some lower level primitives when reading/writing files in the test environment context.

(basically I'm thinking we send around strings, then we create buffer objects for creation, or convert from the buffer object when reading and send back the file contents. I can help with this, but it will take some work.

lkwdwrd (approved):
Approving, but ask for a method rename. In addition, let's pull these tests and come back to them when we can work out how to deal with the cross-platform issues that will arise.

PR #50, opened 2021-02-20
Id458/add username to password reset

rosasck:
I will have to find a way to test this since im unsure how to currently test this but! right above the code i wrote they set the name field for the email_eacp object but they don't set it in the template fields... but the object they are writting into "name" does not exist...  https://github.com/tozny/storage-service/blob/c06564d2882d4f33104f5e208b37bfb157da4972/dao/eacp_dao.go#L41

but this code will trigger the "not empty payload" in here! which is called from this method
https://github.com/tozny/storage-service/blob/59cb28ee59f03fbdf29924b996ce21b9a49d27d8/service/extended_access_control_polices.go#L317

Side note:
not sure if the code above this setting the name is a JS feature, if it is i can switch my line as well


lkwdwrd (approved):

rosasck:
Just added the fix for name in template!

lkwdwrd (approved):

lkwdwrd:
Thank you!

efabens (approved):

PR #51, opened 2021-02-25
Plat630/private realm info endpoint

rosasck:
Im having issues getting the register realm method to work... but here is the endpoint


__tests__/identity.test.js
```diff
@@ -1,64 +1,70 @@
-// const { v4: uuidv4 } = require('uuid')
-// const { apiUrl, clientRegistrationToken, idRealmName, idAppName } = global
-// const Tozny = require('../node')
-// const ops = require('./utils/operations')
+const { v4: uuidv4 } = require('uuid')
+const { apiUrl, idRealmName, idAppName, clientRegistrationToken } = global
```
lkwdwrd, line 2: I think this token needs to get generated with the ability to add 'identity' client types, and I'm not sure off the top of my head if 'normal' tokens in the dashboard allow it... Any token created during realm creation would work for both identity and general clients, though.

lkwdwrd (approved):

efabens (approved):

lib/identity/client.js
```diff
@@ -328,6 +328,24 @@ class Client extends PartialClient {
...
     return await validateResponseAsJSON(response)
   }
+
+  /**
+   * privateRealmInfo fetches private information about a realm
+   *
+   * @return {Object} information about a realm
+   */
+  async privateRealmInfo() {
+    let response = await this.storage.authenticator.tsv1Fetch(
```
lkwdwrd, line 338: It's worth considering adding a cache for this info... If you look at how the basic `realmInfo()` method works, there is an in memory cache that is checked first and returned if the data has already been fetched once. There is also an invalidation flag allowing the user to forcibly fetch fresh data from the server.

See: https://github.com/tozny/js-sdk/blob/3c6d548fc4345569328d01752ab9f1812c43304f/lib/identity/realm.js#L149

For this, the cache should be scoped to the identity object, not globally as the public info cache is (that one uses an array stored globally in the module, this would be better stored as an internal object property).

lkwdwrd (approved):

galxy25 (approved):

PR #52, opened 2021-03-02
Listing shared secrets 

rosasck:
Not sure if this is the correct base branch, i cant figure out which it is and with everyother one it gets Alannah's Commits


lib/identity/client.js
```diff
@@ -172,6 +172,42 @@ class Client extends PartialClient {
     let resultQuery = await this.storage.search(request)
     return resultQuery
   }
+
+  /**
+   * getSharedSecrets
```
lkwdwrd, line 177: Please complete the doc comment details

lib/identity/client.js
```diff
@@ -172,6 +172,42 @@ class Client extends PartialClient {
...
+   *
+   *
+   * @return {Promise<Record>}
+   */
+  async getSharedSecrets() {
```
lkwdwrd, line 182: Does this need it's own method or can we think about adding filters to the list method?

lkwdwrd (commented):
The method looks OK, but this feels like a very specialized method. Noting there will likely be alternate fetches going forward, I think this will grow better over time by modifying your list method to allow filtering more similar to our search endpoints. The logic can be in JS, but eventually I'd like it to more to the server.

Also, let me know if I'm just misunderstanding the user case, which is quite possible!

lib/identity/client.js
galxy25: agree with where your mind is heading, @lkwdwrd @rosasck we need to chat about this before we merge this code in

PR #53, opened 2021-03-02
endpoint support and tests for search identities

rosasck:
Luke mentioned helper methods for a single username/email look up! let me know if thats something I should do!
This will fail jenkins tests until new endpoint is deployed :)


rosasck:
currently adding helper methods

rosasck:
Added Helper methods for single look ups not batch!

PR #55, opened 2021-03-09
List Shared Secrets with me 

rosasck:
@carrala i added them here!


carrala (approved):

lib/identity/client.js
```diff
@@ -209,6 +209,41 @@ class Client extends PartialClient {
     }
     return { exists: false, results: null }
   }
+  /**
+   * getSharedSecrets
+   *
+   *
+   * @return {Promise<Record>}
+   */
+  async getSharedSecrets() {
```
galxy25, line 218: If this isn't needed for the current set of functionality that is deployed or is about to be deployed, probably best to remove this from this pr, otherwise if this is needing to be merged please fill out the documentation for this method

galxy25 (approved):

rosasck:
This will fail until we deploy the new endpoint!

PR #56, opened 2021-03-10
Update dependencies and switch to cross-fetch

lkwdwrd:


lib/storage/index.js
```diff
@@ -10,7 +10,7 @@ const {
   ClientDetails,
   NoteOptions,
 } = require('../../types')
-const fetch = require('isomorphic-fetch')
+const fetch = require('cross-fetch')
```
galxy25, line 13: If you have to make this update in 5 places today might be worth consolidating this behind a Tozny wrapper for fetch

galxy25 (approved):

rosasck (approved):
Would it be wanted/needed to add new browsers? Like edge?

lkwdwrd:
When we fully support edge, yes, but for now it's not on our official support matrix. Since they switched to chromium a while back it should now work fine, but I haven't spun up a VM to check it yet.

galxy25:
> When we fully support edge, yes, but for now it's not on our official support matrix. Since they switched to chromium a while back it should now work fine, but I haven't spun up a VM to check it yet.

@lkwdwrd no VM needed, I just tested TozId login and secret creation, sharing and viewing in native Microsoft edge on my Mac,  all green.

lkwdwrd:
well... let me throw edge latest into the matrix, then! :)

lib/storage/index.js
lkwdwrd: Going to leave as is for the time being. Two reasons:

1. This is a full dependency change because isomorphic-fetch stopped receiving updates. This is very rare in general, but these statements are more like importing golang packages into several files.
2. When we switch to typescript we'll switch to ES6 imports, so much of this import code will get a refactor at that point.

PR #58, opened 2021-03-23
Delete a version of a secret

carrala:


rosasck (approved):

package.json
```diff
@@ -47,7 +47,7 @@
     "eslint-config-prettier": "^8.1.0",
     "eslint-plugin-jest": "^24.2.0",
     "eslint-plugin-prettier": "^3.3.1",
-    "husky": "^5.1.3",
+    "husky": "^5.2.0",
```
rosasck, line 50: not again!!

rosasck (approved):

lib/identity/client.js
```diff
@@ -766,6 +766,51 @@ class Client extends PartialClient {
...
   }
+  /**
+   * removeSingleSecret
+   * @param {String} secretID
+   * @param {String} version
+   * @param {String} secretName
+   * @param {String} secretType
+   * @return {Boolean} returns true if the secret was successfully removed
+   */
+  async removeSingleSecret(secretID, version, secretName, secretType) {
```
efabens, line 777: From a usability perspective I would pass in a secret and then get those values in this method. instead of having 4 parameters that all come from the same object

galxy25: +1

__tests__/secret.test.js
```diff
@@ -570,6 +570,18 @@ describe('Tozny identity client', () => {
     )
     expect(sharedList[0].meta.recordId).toBe(recordView.meta.recordId)
   })
+  it('can delete an unshared secret', async () => {
```
galxy25, line 573: but it does work for shared secrets as well?

lib/identity/client.js
carrala: will do! i got so excited that it was finally working that I completely forgot we talked about this!

__tests__/secret.test.js
carrala: another test is coming for shared secrets -- but yes it does work for them!

lib/identity/client.js
galxy25, line 777: keep the naming consistent, its `createSecret` so I expect `deleteSecret`, remove would be congruent if there was an add method

galxy25: and for single, maybe `deleteSecretVersion`

efabens: maybe even just the secret meta.

efabens: only pushback here is at some point we do need to differentiate between a secret version, and an entire secret (all versions). Which get removeSecret and which has a different name?

lib/identity/client.js
```diff
@@ -766,6 +766,51 @@ class Client extends PartialClient {
...
+          let recordType = `tozny.secret.${SECRET_UUID}.${secretType}.${secretName}`
+          await this.storage.revokeRecordWithGroup(
+            groupInfo.groupID,
+            recordType
+          )
+          await this.storage.deleteGroup(groupInfo.groupID)
+        }
+      })
+    )
+    const groupName = `tozny.secret.${this.config.realmName}.${
```
galxy25, line 797: almost all of this code is a duplicate of what's in lines 782, which is a clear sign it should be extracted out into a re-usable method

lib/identity/client.js
efabens: clearly didn't refresh my page and see you had already said what I repeated

lib/identity/client.js
```diff
@@ -766,6 +766,51 @@ class Client extends PartialClient {
...
+   * @param {String} version
+   * @param {String} secretName
+   * @param {String} secretType
+   * @return {Boolean} returns true if the secret was successfully removed
+   */
+  async removeSingleSecret(secretID, version, secretName, secretType) {
+    const sharedList = await this.getSecretSharedList(secretName, secretType)
+    await Promise.all(
+      sharedList.map(async users => {
+        let identity = await this.searchIdentityByUsername(users.username)
```
galxy25, line 781: @rosasck this is a bulk endpoint right? We should be able to do this search once around 778 to get all the usernames and not have to make a search for each user the secret is shared with.

galxy25 (changes requested):
Functionality wise great work! Needs some refactoring to be a better base for future development

lib/identity/client.js
galxy25: ah ah, don't cross abstractions if it can be avoided, even if its the easiest path forward at the moment. Conceptually secrets exists on top of records (hence why we refer to secretID and not recordID), our "public" sdk methods to work with secrets should accept secret objects, and pull out the lower level bits as needed to accomplish their tasks. If this was an e3db-clients-go lower level library sort of thing sure, but this is an SDK.

lib/identity/client.js
rosasck: `searchIdentitiesByUsername` is a bulk endpoint! 

@carrala  should have caught this during our debug session! we just need a list of all usernames and then call `searchIdentitiesByUsername` NOT `searchIdentityByUsername`

rosasck: in hindsight maybe too close of a name....

carrala:
additional tests for deleting versions of secrets + refactoring!

galxy25 (approved):

efabens (approved):

PR #61, opened 2021-04-23
Feature/refresh

lkwdwrd:
Intial implementation for PKCE flow on login and refresh tokens for sessions. Once the server-side is complete, I will add some integration tests for this login, and validate refreshes are working as expected.


browser/sodium/sodium.js
```diff
@@ -166,9 +166,18 @@ class SodiumCrypto extends CryptoProvider {
     return sodium.crypto_sign_verify_detached(signature, message, publicKey)
   }
 
-  async hash(message) {
-    await sodium.ready
-    return sodium.crypto_generichash(sodium.crypto_generichash_BYTES, message)
+  async hash(message, algorithm = 'Blake2B') {
+    if (algorithm === 'Blake2B') {
```
galxy25, line 170: should we have a constant for `Blake2B` (or does sodium provide one?)

galxy25 (approved):

efabens (approved):

browser/sodium/sodium.js
lkwdwrd: Sodium doesn't. We could make it a constant, but I don't seen it mattering all that much. This is the default for sodium only, if we support other crypto systems in the future, those systems will have an alternate default hash, so if a constant is chosen later on, just make sure we define it and the defaults in the sodium specific files.

PR #63, opened 2021-05-03
secret refresh with namespaces

rosasck:
Will Ask during standup-- 
* Currently our portal for user to user sharing relies on the username to be sent from the backend and we currently are storing that in the group description for that user to user sharing. This is also needed for removing a secret. Here are my two ideas to either rework or keep the methods
A) keep ShareSecretWithUsername which follows the current namespace rules and only use that until we rework some dependent methods 
B) Rework how we keep track of username shared with and removing the share from a user. Currently we are able to get all groupMembers for a group, and we could call this for every secret we have and then look up the clientID and get the username!


lib/identity/client.js
```diff
@@ -886,9 +1096,7 @@ class Client extends PartialClient {
       )
     }
     // if the admin sharing group will be empty, revoke the share & delete the group
-    const groupName = `tozny.secret.${this.config.realmName}.${
-      this.storage.config.clientId
-    }.${secret.meta.plain.secretType.toLowerCase()}`
+    const groupName = `tozny.secret.${this.storage.config.clientId}.${this.config.realmName}`
```
galxy25, line 1099: I think RealmName should still come before clientID as namespacing is usually from general -> less general, and realm (currently) is higher order than the client id.

rosasck: Luke had wanted to switch it for admin namespace incase the user gave their clientID for a namespace there was no way of them accidentally messing with this namespace

lib/identity/client.js
```diff
@@ -366,6 +365,14 @@ class Client extends PartialClient {
     let sharedRecordList = []
     const lengthGroup = groupList.groups.length
     for (var index = lengthGroup - 1; index >= 0; index--) {
+      var groupNameSplit = groupList.groups[index].groupName.split('.')
```
galxy25, line 368: extract this bit of validation logic out to a helper function `isValidToznySecretNamespace`

lib/identity/client.js
```diff
@@ -606,7 +618,117 @@ class Client extends PartialClient {
...
+        usernameLowerCase
+      ) {
+        clientID =
+          clientResponse.searched_identities_information[index].client_id
+      }
+    }
+    if (clientID == null) {
+      return false
+    }
+    // Check if namespace exists
```
galxy25, line 709: this logic of checking for if the namespace exists and if not creating it is very similar to the above, please extract out what is common to a helper function, and then call that in both places, e.g.  `let group = getOrCreateNamespace(namespace)`

lib/identity/client.js
```diff
@@ -602,11 +606,83 @@ class Client extends PartialClient {
...
+  async addSecretToNamespace(secretName, secretType, namespace) {
+    // Check if the namespace exists
+    let currentGroup = await this.getOrCreateNamespace(namespace)
+    // Generate the Record Type to share
+    const recordType = `tozny.secret.${SECRET_UUID}.${secretType}.${secretName}`
+    let sharedWithGroup = await this.storage.shareRecordWithGroup(
+      currentGroup.groupID,
+      recordType
+    )
+    return sharedWithGroup.record_type
```
galxy25, line 629: Very elegant 👏🏽

lib/identity/client.js
```diff
@@ -602,11 +606,83 @@ class Client extends PartialClient {
...
+    if (usernameToAdd == '') {
+      throw new Error('Username Required')
+    }
+    // Look Up the ClientID for Username
+    let clientResponse = await this.searchRealmIdentitiesByUsername([
+      usernameToAdd,
+    ])
+    let usernameLowerCase = usernameToAdd.toLowerCase()
+    let clientID
+    if (clientResponse.searched_identities_information === null) {
```
galxy25, line 650: ideally the response type should handle setting some reasonable default value like the empty array  to avoid having to write this validation logic in our or customers code that uses the sdk

lib/identity/client.js
```diff
@@ -694,13 +770,88 @@ class Client extends PartialClient {
...
+    // Look Up the ClientID for Username
+    let clientResponse = await this.searchRealmIdentitiesByUsername([
+      userToRevokeShare,
+    ])
+    let usernameLowerCase = userToRevokeShare.toLowerCase()
+    let clientID
+    if (clientResponse.searched_identities_information == null) {
+      return null
+    }
+    for (
```
galxy25, line 799: this code looks familiar, consider extracting

lib/identity/client.js
galxy25: pending impactful examples and with the assumption that any way things could get messed with in regards to client id is the same liability for messing with things by providing realm name as part of the namespace, please change this back to

`{tozny.secret}.{realmName}.{namespace} where namespace could be clientID`

galxy25 (approved):
None of my comments are blocking, I'm looking the other way on explicit sleeps in test code because its test code, but if those tests keep failing and its due to time outs, take the time to write some utility test helpers to more resiliently waiting for some group state  to be true / ready

lib/identity/client.js
```diff
@@ -366,6 +341,9 @@ class Client extends PartialClient {
     let sharedRecordList = []
     const lengthGroup = groupList.groups.length
     for (var index = lengthGroup - 1; index >= 0; index--) {
+      if (!this.isValidToznySecretNamespace(groupList.groups[0].groupName)) {
```
lkwdwrd, line 344: thanks for adding this!

lib/identity/client.js
```diff
@@ -602,11 +562,63 @@ class Client extends PartialClient {
...
+    if (usernameToAdd == '') {
+      throw new Error('Username Required')
+    }
+    // Look Up the ClientID for Username
+    let clientID = await this.clientIDLookUpByUsername(usernameToAdd)
+    if (clientID === null) {
+      return null
+    }
+    // Check if namespace exists
+    let currentGroup = await this.getNamespace(namespace)
```
lkwdwrd, line 606: This should also run the getOrCreate I think -- basically if you add an identity or a record to a namespace, it will cause it to exist.

This is similar to the SDKs for sharing where if you share a type before you write any records of that type, it will first generate an AK for the tuple, and then do the share. When records are written to that type in the future, the share is active.

lib/identity/client.js
```diff
@@ -694,13 +686,66 @@ class Client extends PartialClient {
...
+      return null
+    }
+    // Look up group for our current namespace
+    let currentGroup = await this.getNamespace(namespace)
+    // If namespace doesn't exist return true
+    if (currentGroup.groupID === undefined) {
+      return true
+    }
+    // Remove identity from namespace
+    let revokeFromNamespace = await this.storage.removeGroupMembers(
```
lkwdwrd, line 718: In JS, instead of awaiting into a variable, and then immediately returning that variable, just return the function call. It doesn't matter much, but the difference can add up in wrapper land.

return as written:

- extra variable memory location to point to the promise from removeGroupMembers
- awaiting at line 718 makes the return Promise<Promise<removeGroupMemebers()>>

return directly
- async determines if the return is plain or a promis, and wraps the plain return in a promise as needed
- `return this.storage.removeGroupMembers(...)`
- Return is Promise<removeGroupMemebers()>

lib/identity/client.js
```diff
@@ -694,13 +686,66 @@ class Client extends PartialClient {
...
+   * @param {String} secretType The secret type, example 'Credential'
+   * @param {String} secretName The user given name for the secret stored in the meta data
+   *
+   * @return {Boolean}
+   */
+  async removeSecretFromNamespace(secretName, secretType, namespace) {
+    // Look up group for our current namespace
+    let currentGroup = await this.getNamespace(namespace)
+    // If namespace doesn't exist return true
+    if (currentGroup.groupID === undefined) {
```
lkwdwrd, line 738: bailing early on remove makes a ton of sense here 👍

lib/identity/client.js
```diff
@@ -886,16 +926,99 @@ class Client extends PartialClient {
...
       secret.meta.recordId,
       secret.meta.version
     )
   }
+  /**
+   * isValidToznySecretNamespace
+   * @param {Object} groupName
+   * @return {Boolean} returns true if the secret was successfully removed
+   */
+  async isValidToznySecretNamespace(groupName) {
```
lkwdwrd, line 942: This is probably not useful in the client context (it's not something I would ever expect a user to want to call) for that reason, I'd like to move it to a utility function instead.

lib/identity/client.js
```diff
@@ -886,16 +926,99 @@ class Client extends PartialClient {
...
+      return false
+    }
+    return true
+  }
+  /**
+   * getNamespace
+   * @param {Object} namespace
+   * @return {Object} returns the group based on namespace
+   */
+  async getNamespace(namespace) {
```
lkwdwrd, line 958: Because the work is so similar, I think this and the `getOrCreateNamesapce` methods can be combined. Also, this is public on the client, but it is not a function we want users to know or think about really. Namespace is an abstract concept. So for that reason I may pull this out as a protected function defined above the class. Keep the consumer interface a little cleaner with less lower level footguns available.

lkwdwrd (changes requested):
The functionality looks spot on. There are some code organizational things I'd like to see cleaned up before publish, but should mostly be moving things around to clean up the consumer interface for our identity clients.

lib/identity/client.js
```diff
@@ -886,16 +938,72 @@ class Client extends PartialClient {
...
       secret.meta.recordId,
       secret.meta.version
     )
   }
+  /**
+   * getOrCreateNamespace
+   * @param {Object} namespace
+   * @return {Object} returns the group based on namespace
+   */
+  async getOrCreateNamespace(namespace) {
```
lkwdwrd, line 954: This still needs to get refactored out into a private/protected method (removed from living within the client class itself).

lib/identity/client.js
```diff
@@ -602,11 +561,69 @@ class Client extends PartialClient {
...
+    if (clientID === null) {
+      return null
+    }
+    // Check if namespace exists
+    const groupName = `tozny.secret.${this.config.realmName}.${namespace}`
+    const groupList = await this.storage.listGroups(
+      this.storage.config.clientId,
+      [groupName]
+    )
+    if (groupList.groups.length < 1) {
```
lkwdwrd, line 610: DId the `getNamespace` method get refactored out?

This method should do the same `getOrCreateNamespace` but I'd like to adjust all of it to a single helper method which can perform the `getNamespace` or `getOrCreateNamescpace` based on call params:

```ts
function getNamespace(namespace: string, createIfMIssing: boolean){
...
}
```

This kind of signature.

lkwdwrd (changes requested):
A couple things I think I didn't explain well enough. Let me know if this this is confusing and we can discuss in real time.

lkwdwrd (approved):

efabens (approved):

PR #66, opened 2021-05-27
 Actually fix serialization of plain search term.

galxy25:
Tests for happy and sad paths.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@toznysecure/sdk",
-  "version": "1.3.1-alpha.2",
+  "version": "1.3.3",
```
rosasck, line 3: `1.3.2`

rosasck (approved):
Should it be 1.3.2?

package.json
rosasck: ignore this! just saw there was another publish

PR #67, opened 2021-05-27
Fix secret deletion

carrala:
Things that have changed:
- createSecret only shares the record type with the group if it's a new secret. This prevents multiple keys being made for the same record type.
- listRecordsSharedWithGroup is called within a try catch block & a processing error is recorded if it fails within listing groups. This ensures that listing secrets is possible even if a group is inaccessible.
- getSecretSharedList returns all the usernames a secret is shared with, but does not return the username of the secret owner, which makes delete more efficient & displays the correct usernames in the identity portal.
- deleteSecretVersion handles multiple types and versions of secrets shared within a group and deletes the personal sharing group if it's empty.
- the new test makes sure that all of the secrets can be deleted without causing any errors.


lib/identity/client.js
```diff
@@ -332,9 +336,14 @@ class Client extends PartialClient {
...
-      let sharedRecords = await this.storage.listRecordsSharedWithGroup(
-        groupList.groups[index].groupID
-      )
+      let sharedRecords = []
+      try {
+        sharedRecords = await this.storage.listRecordsSharedWithGroup(
+          groupList.groups[index].groupID
+        )
+      } catch (e) {
+        continue
```
galxy25, line 345: at the very least we should log this error, at the most we should collect them and return it as part of the response from this function

lib/identity/client.js
```diff
@@ -428,7 +462,11 @@ class Client extends PartialClient {
         }
       }
     }
-    return { list: sharedRecordList, nextToken: groupList.nextToken }
+    return {
+      list: sharedRecordList,
+      nextToken: groupList.nextToken,
+      processingErrs: processingErrs,
```
galxy25, line 468: `processingErrors`

lib/identity/client.js
```diff
@@ -952,13 +1063,21 @@ class Client extends PartialClient {
...
+            secret.meta.type,
+            secret.meta.recordId
+          )
         })
       )
     }
     // if the admin sharing group will be empty, revoke the share & delete the group
-    const groupName = `tozny.secret.${this.storage.config.clientId}.${this.config.realmName}`
-    await this.revokeShareBeforeDeleteSecret(groupName, secret.meta.type)
+    const groupName = `tozny.secret.${this.config.realmName}.${this.storage.config.clientId}`
```
galxy25, line 1075: do we have a method for this, if not can we create one please?

galxy25 (approved):
Approved with non-blocking comments

lib/identity/client.js
```diff
@@ -793,16 +862,29 @@ class Client extends PartialClient {
...
-          }) // add their username to the list which is stored in the group description
+          if (
+            this.validGroupForSecretSharedList(
+              groupList.groups[index],
+              secretName,
+              secretType
+            )
+          ) {
+            groupSharedList.push({
+              username: groupList.groups[index].description,
```
rosasck, line 884: This only exists if its between a user to user not for a namespace since we add to the namespace we dont create a new group between the users

rosasck (commented):
Can you verify this works for namespaces as well

__tests__/secret.test.js
```diff
@@ -680,13 +678,25 @@ describe('Tozny identity client', () => {
       await new Promise((r) => setTimeout(r, 200))
     }
     let sharedList = await ops.getSharedSecrets(realmConfig, identity2)
-    expect(sharedList[0].data.secretValue).toBe('secret-value')
+    expect(sharedList.sharedList[0].data.secretValue).toBe('secret-value')
```
rosasck, line 681: will this break the identity portal?

rosasck (approved):
pending portal question

PR #71, opened 2021-07-21
Update package.json and build output to be in sync with the latest published version 1.3.6

galxy25:


DEVELOPMENT.md
```diff
@@ -190,7 +190,7 @@ npm publish --tag=alpha
 Push the tag up to remote github repository
 
 ```bash
-git push --tags --all
+git push origin vX.Y.Z # e.g. git push origin v1.0.1
```
efabens, line 193: While this requires explicitness it also causes an increased likelihood of user error. To some extent we are weighing does a user have unrelated tags on their local box vs do we think a user will fat finger a tag name. I would lean more towards if a user has unrelated tags on the local they probably have a flow that can account for that and we shouldn't need to remember the tag we are pushing

efabens: But I am not strongly opinionated, https://stackabuse.com/git-push-tags-to-a-remote-repo this person disagrees with me and makes valid points.

efabens (approved):

DEVELOPMENT.md
galxy25: added an environment variable export step to minimize (or at least consistenize any typoing) chances for human error during the publishing process

PR #73, opened 2021-10-04
[WASH-1059] Identity locked login request error

Mrhea:
Putting this up as a draft PR because I have not tested it yet.


galxy25 (approved):

carrala (approved):

rosasck (approved):
whoop whoop!!

house9:
Should we be concerned about the changes in `package-lock.json`, it looks like it downgrades from `lockfileVersion` 2 to 1.

I am guessing this was needed to get it running locally 

https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json#lockfileversion

> Note that the file format changed significantly in npm v7 to track information that would have otherwise required looking in node_modules or the npm registry.

I guess this is why that file is massive when using `lockfileVersion` 2 

I checked on my local machine by switching node versions and you can see corresponding stock npm versions.
But I think you can probably install newer versions of npm on your existing node version? Or use [asdf](https://github.com/tozny/internal-docs/blob/master/welcome/asdf.md) or nvm to support multiple versions on your local machine

```
> asdf local nodejs 14.17.5
> node -v
v14.17.5
> npm -v
6.14.14

> asdf local nodejs 16.6.2
> node -v
v16.6.2
> npm -v
7.21.1
```

lib/utils/index.js
```diff
@@ -83,9 +83,15 @@ async function credentialedDecodeResponse(response) {
...
-      throw new errors.identity.InvalidCredentials(e.response, e.message)
+    if (e instanceof errors.general.RequestError) {
+      if (e.response.status === 401) {
+      // If we received unauthorized, map it to an invalid credentials error.
+        throw new errors.identity.InvalidCredentials(e.response, e.message)
+      }
+      if (e.response.status === 417) {
+      // If we received expectation failed, map it to an identity locked error.
+        throw new errors.identity.IdentityLockedError(e.response, e.message)
+      }
```
pirtleshell, line 94: for readability & a future world where we handle 418s or other error codes, i'd consider consolidating these `if` blocks into a `switch-case`
```js
switch (e.response.status) {
  case 401: // ...
  case 417: // ...
  default: break
}
```

package-lock.json
```diff
@@ -1,8595 +1,8 @@
 {
   "name": "@toznysecure/sdk",
   "version": "1.3.6",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
pirtleshell, line 4: as @house9 alludes to, in npm v7 they changed how lockfiles are structured.

can you regenerate this file with an upgraded `npm` installed to prevent downgrading?
```sh
# upgrade npm
npm upgrade -g npm@latest
# regenerate package-lock
npm i
```

pirtleshell (approved):
approved pending regeneration of package-lock

package-lock.json
pirtleshell: (insert a @galxy25 rant on the benefits of containerization 😆) 
(though to be fair, one is rarely using `npm` to add new packages from the container :trollface:)

lib/identity/partialClient.js
```diff
@@ -54,25 +55,34 @@ class PartialClient extends CryptoConsumer {
...
-      await checkStatus(request)
-      return true
-    } catch (_) {
-      return false
+      await checkStatus(response)
+
+      ret = {success: true, message: ""}
+      return ret
+    } catch (e) {
+      switch (e.response.status) {
```
Mrhea, line 71: The Identity Portal update password flow does not check for any error objects, so in order to make minimal changes we check the possible cases here.

lib/identity/partialClient.js
```diff
@@ -54,25 +55,33 @@ class PartialClient extends CryptoConsumer {
...
           body: JSON.stringify(body),
         }
       )
-      await checkStatus(request)
-      return true
-    } catch (_) {
-      return false
+      await checkStatus(response)
+      ret = {success: true, message: ""}
+      return ret
```
pirtleshell, line 68: unlike golang (😅), prefer immutability in js. you can just return the value directly.

lib/identity/partialClient.js
pirtleshell: you should check that `e.response` exists before this switch and return false with no message

lib/identity/partialClient.js
```diff
@@ -54,25 +55,33 @@ class PartialClient extends CryptoConsumer {
...
+    } catch (e) {
+      switch (e.response.status) {
+        case 401:
+          ret = {success: false, message: "Current password is incorrect."}
+          break
+        case 417:
+          ret = {success: false, message: "Your Identity is currently locked due to too many failed login attempts. Please try again later."}
+      }
+      return ret
```
pirtleshell, line 77: default `ret` to `{success: false}` at the start of this `catch` block

pirtleshell (changes requested):

lib/identity/partialClient.js
```diff
@@ -54,25 +54,36 @@ class PartialClient extends CryptoConsumer {
...
         }
       )
-      await checkStatus(request)
-      return true
-    } catch (_) {
-      return false
+      await checkStatus(response)
+      return {success: true}
+    } catch (e) {
+      let ret = {success: false}
```
pirtleshell, line 68: on second glance, there should also be a default message that says something like "something went wrong" as a catchall for any other errors that may happen. otherwise we swallow the error with no indication that it didn't go as planned

the switch-case below can just set the `message` like `ret.message = "Current password is incorrect"`

pirtleshell (approved):
approved pending a default message passed to the user on non 401 or 417

lib/identity/partialClient.js
Mrhea: Ah! My beginner Javascript skills are showing

Mrhea: Would it be too much to do something like `Something went wrong. ${e}`

pirtleshell: you'd want `e.message`, i think. dealers choice. i tend to shy away from displaying potentially confusing errors to users 😆

PR #76, opened 2021-10-14
[WR-173] implement approveAccessRequests

house9:


lib/identity/client.js
```diff
@@ -1252,6 +1254,29 @@ class Client extends PartialClient {
     await checkStatus(response)
     return { success: true }
   }
+
+  /**
+     * approveAccessRequests approves one or more access requests
+     *
+     * @param {Object} AccessRequestApprovalsRequest
```
pirtleshell, line 1261: `@param` should describe the arguments to this func. can you document that `realmName` is a string & that `approvals` is an array of objects with an id & comment?

i care less about defining an actual `{Object}` & more about just something you could quickly read and know what to pass in

pirtleshell (approved):

PR #77, opened 2021-10-15
[WR-175] change search access request signature

pirtleshell:
* changes multiple filter array arguments to be a single filters object
* ~~adds the :sparkles: new :sparkles: `approvableByMe` filter option~~

~~`approvableByMe` is forthcoming (soon enough to check it in here now), but does not affect calls to this endpoint for the currently supported filters.~~

we opted against an `approvableByMe` filter.


.env.example
```diff
@@ -5,4 +5,4 @@ API_URL=https://dev.e3db.com
 CLIENT_REGISTRATION_TOKEN=<existing_client_token_here>
 ID_REALM_NAME=<existing_realm_name_here>
 ID_APP_NAME=account
-TEST_TOZID_GROUP_ID=<existing_group_id_here>
+TEST_TOZID_GROUP_ID=<existing_group_id_that_has_access_policy_here>
```
house9, line 8: 👍

rosasck (approved):

PR #78, opened 2021-10-15
[WR-173] refactor approval response

pirtleshell:
the `approveAccessRequests` function was decoding to a class that had no
constructor. this means that no matter the response data, the client func
could only return and empty object. fixing that, it is more user friendly
to just hand back the list of updated access requests.


galxy25 (approved):

rosasck:
should be none

PR #80, opened 2021-10-18
[WR-225] add requestor username to accessRequest

house9:
See: https://github.com/tozny/identity-management-service/pull/256


house9:
All my js tests are currently failing locally (with or without this change) - going to rebuild my local environment to confirm this had no adverse effects.

PR #88, opened 2021-11-01
[WR191] First login Bug found

rosasck:
@carrala found a great bug in the first time login!

Bug: First time login would get a 401, every other login succeeds.

Reason: User doesn't exist when creating the challenge, we need to restart login after registering to be able to validate the challenge and password!


galxy25 (approved):
Appropriately humble PR title!

carrala (approved):

sjonz (approved):

lib/identity/realm.js
```diff
@@ -398,6 +398,10 @@ class Realm extends CryptoConsumer {
         false,
         60
       )
+      // Upon first call to intiate login. we cannot create a challenge for a user that
+      // isn't registered. After Registration we must begin the login flow once again
+      // to create the challenge correctly for the user or we get a 401
+      return await this.login(username, password, actionHandler, credentialType)
```
pirtleshell, line 404: this `await` is unnecessary

PR #92, opened 2021-11-12
[WR-262] add access policies to access request groups

house9:


house9:
Was planning to build `dist` along with adding the new version on a separate PR, but can do it as part of this PR instead?

rosasck (approved):

rosasck:
> Was planning to build `dist` along with adding the new version on a separate PR, but can do it as part of this PR instead?

That was really easy to read without the dist

galxy25 (approved):

PR #97, opened 2022-01-28
[PLAT-862] fix broken test

pirtleshell:
you can't currently approve your own requests, so the approve test needs
a second identity to perform the operation.

hopefully this fixes the broken CI pipeline.


__tests__/identity.test.js
```diff
@@ -267,7 +276,9 @@ describe('Tozny identity client', () => {
     }
     const response = await ops.approveAccessRequests(
       realmConfig,
-      identity,
+      // you can't approve your own requests so a different identity must perform approval.
```
galxy25, line 279: this is a desired feature though and would immediately make everyone's life at Tozny easier that has to go through mpc flows for things that we can trust (but want to audit) tozny engineers to be able to self approve for, I would strongly consider "just be a hero (a super hero)" if adding support for this feature is less than a days work.

galxy25 (approved):

pirtleshell:
after a lot of digging, i've discovered that our `operations` pattern for running code in different environments does support native classes like `Date`. the failing `SearchRange` test is the only place such classes are used. 

rather than remove the `instanceof` type checking... i made it a feature. the `SearchRange` class now supports any input for `start` and `end` that can be parsed by javascript's `Date.parse` func. this means that it will continue to accept `Date`s, but will also accept strings like `"2022-01-28"`, and even the strange `Date.prototype` objects that are generated by our cross platform process of converting funcs to strings that are then injected into another string func that is converted to a `Function` that is `.apply()`-ed 🌀 🤪 

i should note this is not a problem with how the test was originally written as suspected by @galxy25. the operations func handles converting the serialized dates back to date objects... for some reason they are just different instances of the Date class than the browser's in firefox and chrome (but not Safari....)

tl;dr: `SearchRange` is more permissive in what arguments it allows for `start` and `end` dates, while still preserving the runtime typecheck validation.

efabens (approved):

pirtleshell:
✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ ✅ 
Success-kid

PR #117, opened 2022-07-21
Updates README to include required values to accessing a realm

lily-tsou:


README.md
```diff
@@ -775,10 +775,12 @@ If there is extra data that is required for the computation to run, it must be i
...
 ```js
-const realmName = '...'
-const appName = '...'
+const realmName = 'NameOfRealmGoesHere'
+// Currently 'account' is the only app name supported value
+const appName = 'account'
 // This is a URL in your application which handle password reset flows.
-const brokerTargetURL = '...'
+// The default value using the Tozny Hosted Broker is https://<TozIdBaseURL>/<YourRealmName>/recover
+const brokerTargetURL = 'https://id.tozny.com/NameOfRealmGoesHere/recover'
```
tozny-shakeeb, line 783: Suggestion: using {realmName} instead of NameOfRealmGoesHere would make it more obvious to use the realm name here

lily-tsou: @tozny-shakeeb Updated!

lily-tsou: @tozny-shakeeb or did you mean to use an example realm name, not literally "realmName"?

PR #149, opened 2023-05-09
Added Realm group Method

Jayanthi-moorthi:
All  methods related TozIdGroup Creations added.

### DIFF
--- a/lib/storage/client.js
+++ b/lib/storage/client.js
@@ -1814,24 +1814,20 @@ class Client extends CryptoConsumer {
     return validateResponseAsJSON(response)
   }
 
-  async updateRealmGroup(realmName,tozId,tozStoreId){
-
-    const payload = {
-      attributes: {tozStoreId: [tozStoreId]}
-
-    }
-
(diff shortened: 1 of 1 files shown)


lib/storage/client.js
```diff
@@ -1796,6 +1796,65 @@ class Client extends CryptoConsumer {
...
+      }
+    )
+
+    return validateResponseAsJSON(response)
+  }
+
+  async updateRealmGroup(realmName,tozId,tozStoreId){
+
+    const payload = {
+      attributes: {tozStoreId: [tozStoreId]}
```
rosasck, line 1820: can we update this to pass in 
`async updateRealmGroup(realmName,groupID,attributes={})`

and to the method pass in the attributes object, since if this is used for anybody else they may not want this

Jayanthi-moorthi: Okay Katie. I update the same

rosasck (approved):

lily-tsou (approved):

PR #164, opened 2023-07-05
New End Point - Get Groups with Capabilities

Jayanthi-moorthi:


Jayanthi-moorthi:
To increase the performance , add end point to fetch information at a time as part of TOZFLOW-138

rosasck (approved):

PR #192, opened 2024-03-18
Plat 1611 fetch group i ds by capability

rosasck:


rosasck (commented):
looks good to me, ping me when Travis tests finish running and I'll approve

PR #224, opened 2025-01-10
Added direct share sort change

pravinkpk30:
Added shared_record_sort flag


lily-tsou:
@pravinkpk30 https://github.com/tozny/e3dbSearchService/pull/124 is this the only dependent PR?

PR #225, opened 2025-02-04
Removed unused varaible

Jayanthi-moorthi:
Description:
Sending a "Shared URL" link to users via email when sharing records with users outside the realm


lily-tsou:
Approved. Some travis tests are failing but I've retriggered

lily-tsou (approved):

PR #229, opened 2025-04-29
Refresh token expire

Jayanthi-moorthi:
When the refresh token is expired, we need to throw session expired. For that, we are checking refresh token against current date.

### DIFF
--- a/lib/identity/client.js
+++ b/lib/identity/client.js
@@ -102,25 +102,30 @@ async function getNamespace(client, namespace, createIfMissing) {
  * @return {Object} The new _agentToken information with absolute expiry dates
  */
 async function refreshAgentToken(client) {
-  const now = Math.floor(Date.now() / 1000)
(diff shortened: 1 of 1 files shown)


lib/identity/client.js
```diff
@@ -169,13 +175,18 @@ class Client extends PartialClient {
...
+    try {
+      const now = Math.floor(Date.now() / 1000)
+      if (!this._agentToken || (this._agentToken.expires && this._agentToken.expires < now) || 
+        (this._agentToken.expiry && Date.now() > new Date(this._agentToken.expiry).getTime())
+      ) {
+        this._agentToken = await refreshAgentToken(this)
+      }
+      return Object.assign({}, this._agentToken)
+    } catch (error) {
+      // Preserve the error and propagate it to the Svelte application
```
karthikpdc, line 187: I think it should be "Preserve the error and propagate it to the application".

karthikpdc (approved):

